PR #23, opened 2026-08-15
Closes #20: pass -w to pnpm for update too (workspace-root profiles)

### LINKED ISSUE #20: Install/update/remove fail on workspace-root profiles: no `-w` + pnpm version mismatch (ERR_PNPM_ADDING_TO_ROOT / ERR_PNPM_PUBLIC_HOIST_PATTERN_DIFF)
# Install/update/remove fail on workspace-root profiles: no `-w` + pnpm version mismatch

> TL;DR（中文）：市场 UI 的安装/更新/卸载在「pnpm workspace root」的 profile 上必然失败——① `runDshPlugin` 传 `add/remove` 不带 `-w`，pnpm 报 `ERR_PNPM_ADDING_TO_ROOT`；② 机器 pnpm（corepack 提权的 11.x）与建 profile 的 pnpm（9.x）默认 hoist 不同，报 `ERR_PNPM_PUBLIC_HOIST_PATTERN_DIFF`。两个都能复现，附 workaround。

## Environment

- dshmarket installed in a `web` profile whose `pnpm-workspace.yaml` contains `packages: - .` (i.e. the profile dir is a pnpm workspace root — this is the default profile template).
- Machine `pnpm` is what `corepack enable pnpm` provisions (currently 11.x); the profile's `node_modules` was built with pnpm 9.x.

## Bug 1 — no `-w` on workspace-root profiles

`lib/routes.js` forwards `add`/`remove` to `dsh plugin` without the `-w` flag:

- install / update: `runDshPlugin(config.profile, ['add', target])` (routes.js:571 / :423)
- remove: `runDshPlugin(config.profile, ['remove', name])` (routes.js:500)

On a workspace-root profile, `pnpm add`/`remove` abort with:

```
ERR_PNPM_ADDING_TO_ROOT — Running this command will add the dependency to the workspace root…
```

The market UI then surfaces only the wrapper text: `dsh: pnpm failed in profile directory …`. So **install, update, and remove via the market UI all fail on the default web profile**.

## Bug 2 — pnpm version mismatch → hoist-pattern diff

After bypassing the `-w` check, the same flow fails with:

```
ERR_PNPM_PUBLIC_HOIST_PATTERN_DIFF — This modules directory was created using a different
public-hoist-pattern value. Run "pnpm install" to recreate the modules directory.
```

The profile's `node_modules/.modules.yaml` records pnpm 9's default `publicHoistPattern: ['*eslint*','*prettier*']`; pnpm 11's built-in default differs, and the check compares against the built-in default (pinning `public-hoist-pattern` in `pnpm-workspace.yaml` did **not** satisfy it — verified). So any add/update also fails when the machine pnpm major differs from the one that built the profile.

## Suggested fixes

1. **Bug 1**: detect the workspace-root case and pass `-w` (or set `ignore-workspace-root-check`), so `add`/`remove`/`update` work on profile templates.
2. **Bug 2**: pin/provision the same pnpm major the profile was built with (or verify `public-hoist-pattern` compatibility and surface a clear "recreate modules directory" message instead of the opaque wrapper error). At minimum, surface pnpm's real stderr to the UI — today only "dsh: pnpm failed in profile directory" is shown, which hides both root causes.

## Verified workaround (no code change)

- CLI: `dsh plugin --profile web add/update -w <pkg>` with the profile's pnpm major;
- or `ignore-workspace-root-check=true` in the profile's `.npmrc` + aligning the machine pnpm via `corepack prepare pnpm@9.15.9 --activate`.

Thanks — the market is a great addition; these two friction points make the UI's install/update unusable on the stock profile though.

xxiaoxiong:
## Summary

Fixes #20. The install/update/remove flow on a **workspace-root profile** (the default  profile template, whose  has ) fails because  only passed  for /, not .

- pnpm 9 refuses // at a workspace root without  → .
- pnpm 10/11 accept the flag as a no-op there, so passing it always is safe.

## Change

```ts
-  if (pluginArgs[0] === 'add' || pluginArgs[0] === 'remove') {
+  if (pluginArgs[0] === 'add' || pluginArgs[0] === 'remove' || pluginArgs[0] === 'update') {
```

## Verification

- `npm run typecheck` ✅ (tsc -p tsconfig.json + tsconfig.client.json)
- `npm run build` ✅
- `node scripts/preflight.mjs` ✅
- `node scripts/smoke-spawn.mjs` ✅

One-line change, the update route goes through the same `runDshPlugin` path as add/remove, so this closes the remaining gap on workspace-root profiles without touching non-workspace flows.

### DIFF
--- a/src/routes.ts
+++ b/src/routes.ts
@@ -176,7 +176,7 @@ function runDshPlugin(profile: string, pluginArgs: string[]): Promise<InstallRes
   const { file, args, cwd, viaShell } = dshArgv()
   // pnpm 9 refuses to add at a workspace root without -w (#17); pnpm 10/11
   // accept the flag as a no-op there, so it is safe to pass always.
-  if (pluginArgs[0] === 'add' || pluginArgs[0] === 'remove') {
+  if (pluginArgs[0] === 'add' || pluginArgs[0] === 'remove' || pluginArgs[0] === 'update') {
     pluginArgs = [pluginArgs[0], '-w', ...pluginArgs.slice(1)]
   }
   const target = pluginArgs[pluginArgs.length - 1] ?? ''


moonquake2004:
Thanks for the fix — this is exactly the gap we filed in #20 (we're the reporters; our repro was on a workspace-root profile, missing `-w` → `ERR_PNPM_ADDING_TO_ROOT`).

Heads-up: the refactor merged to main after this PR was opened (ea47e0de, 'layer the host into focused modules') already supersedes this one-liner in a way that's worth knowing about:

- The update route no longer calls `dsh plugin update` at all — it re-runs `['add', target]` ("Re-running add re-resolves the source"), and `add` goes through `pluginArgsFor` in `src/pnpm-compat.ts`, which injects `-w` **conditionally** — only when the profile has a `pnpm-workspace.yaml`.
- The unconditional `-w` in this PR's diff would regress non-workspace profiles: per the compat matrix in `pnpm-compat.ts`, every pnpm major rejects `-w` when the directory is not a workspace (`--workspace-root may only be used inside a workspace`). `pluginArgsFor` deliberately avoids that by gating on the yaml's existence.

So the workspace-root `update` gap is closed on main via the `add` path, and the conditional approach is the safer pattern. If a `dsh plugin update` path is ever reintroduced, adding `'update'` to `pluginArgsFor`'s condition (same yaml gate) would be the way. Happy to close this PR or help rework it onto the new architecture — either is fine from our side. Verified on our side that pnpm 9.15.9 + the conditional `-w` + `ignore-workspace-root-check` all coexist without breaking install/update/remove.

xxiaoxiong:
关闭：main 上 ea47e0de 重构已经将 update 路径改为重新运行 add，pluginArgsFor 以条件性方式注入 -w，该 PR 不再适用。感谢修复尝试。

PR #31, opened 2026-08-15
chore: refresh registry snapshot from awesome-dsh-plugin.com

luzhengyangtx:
Refreshes data/registry-snapshot.json with the current https://awesome-dsh-plugin.com/plugins.json (the catalog the market serves), using the repo's own snapshot script command (curl plugins.json). This brings in plugins listed since the last refresh, including dsh-telegram-duty.


fkysly:
Verified: parses clean, count matches the field (457), every url is github.com, every npm name passes the package-name allowlist, categories are the known set, and the content is byte-identical in count to the live plugins.json right now. Exactly what `npm run snapshot` produces — thanks for keeping the offline fallback fresh! Merging.

PR #34, opened 2026-08-15
feat(registry): add offline registry validation gate (P1-8)

max257026-svg:
Add scripts/validate-registry.mjs — a dependency-free, offline gate over data/registry-snapshot.json that catches the catalog defects which would otherwise surface as broken installs or ambiguous entries:

  E1  required string fields present
  E2  bilingual description (en + zh)
  E3  npm name syntax
  E4  stars is a non-negative integer
  E5  github repo / tree(subpath) url
  E6  owner matches url owner
  E7  category in whitelist
  E8  awesome-dsh-plugin page url
  E9  added is a valid, non-future date
  E10 install references the real target
  E11 no duplicate install identity (repo+subpath / npm)
  E12 snapshot count integrity

Display-name collisions across distinct authors / monorepos are reported as warnings, not failures (the catalog permits same name under different owners).

Wire it into package.json (validate:registry) and CI so it runs on every PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
       - run: npm run check
       - run: npm test
       - run: node scripts/preflight.mjs
+      - run: node scripts/validate-registry.mjs
       - run: node scripts/smoke-spawn.mjs
 
   # Real-pnpm matrix (9/10/11 via npx) pinning the failure signatures behind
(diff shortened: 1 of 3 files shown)


fkysly:
Test-drove it against the current snapshot: passes clean, and the warnings section immediately surfaced the four display-name collisions in the catalog (including the dsh-pet-style confusion that hit #25) — exactly the informational signal we wanted. Nice judgment calls: warnings never failing the gate, and keeping network probes out of the merge path. Merging — this closes P1-8's offline half. Thanks!

PR #63, opened 2026-08-15
feat: add profile backup and restore with WebDAV

### LINKED ISSUE #55: 备份和恢复功能
希望能添加一个备份和恢复功能，最好支持webdav

mintonight:
## What changed

- add portable profile backup and restore while excluding `node_modules`, lock state, market cache, and old `.bak-*` files
- export format `0.2` as readable structured JSON: `package.json` is a JSON object and text configuration uses line arrays
- validate paths, write atomically, roll back on reinstall failure, and reject missing local `link:` projects with the dependency and path in the error
- add local export/import and WebDAV upload/restore with daily automatic backup and provider presets
- add bilingual UI plus unit and route coverage

## Why

Profiles had no small, portable way to move their plugin list and configuration between machines. The initial Base64 representation was also difficult to inspect, and missing local projects failed too late during dependency installation.

## Impact

Users can inspect exported backups directly, import them locally, or store them on WebDAV. Restores reject invalid local links before overwriting profile files, reinstall dependencies, and ask for a restart. This test-version format intentionally does not restore older Base64 backups.

## Validation

- `npm test -- --run tests/backup.spec.ts tests/flows.spec.ts` — 35 passed
- `npm run typecheck`
- `npm run build`
- live `/dsh-market/backup` response verified on `dsh web`

Closes #55


fkysly (changes requested):
Thorough work on the backup format itself — the path validation (absolute/`..`/symlink/dup rejection, resolve+prefix check, atomic rename, rollback), the size/count caps, and `redirect: 'error'` on WebDAV are all solid. Three blockers before this can land, though:

## 1. The full test suite is red on this branch

`npm test` → **12 failed** (`tests/client/market-section.client.spec.tsx`): every case dies with `TypeError: localStorage.setItem is not a function`. The new auto-backup effect writes `localStorage` unconditionally on mount, and the client-test environment doesn't provide it. The PR's validation ran only `tests/backup.spec.ts tests/flows.spec.ts` — please run the whole suite and guard the storage access (or stub it in the test setup).

## 2. Security concerns (the market runs same-origin with every other plugin's client code)

- **WebDAV credentials in `localStorage`, plaintext** (`dshm-webdav` incl. `password`). Any same-origin script — i.e. any installed plugin's client bundle — can read them. At minimum keep the password server-side (config) or in memory only; storing it browser-side makes the market the softest credential target in the profile.
- **`POST /dsh-market/webdav` is an SSRF primitive**: the server fetches an arbitrary `http(s)` URL with caller-supplied Basic auth, and `action:'restore'` **returns the fetched body to the caller**. Same-origin gating means any plugin's client code can read internal-network HTTP responses through it (`http://169.254.169.254/…`, `http://localhost:…`). Suggest: https-only by default, block private/link-local ranges, and don't echo the body for non-backup payloads (validate as a backup server-side before returning).
- **`GET /dsh-market/backup` has no same-origin/loopback gate and exports profile files unsanitized**. The log-export flow deliberately masks credentials; this endpoint hands out `config.toml` and friends verbatim — API keys included. Please gate it like `/dsh-market/restart` (loopback + same-origin) and consider masking known secret fields, or at least a loud UI warning that backups contain credentials (they'll be uploaded to WebDAV, too).

## 3. Rebase needed

Today's main gained #76/#77/#78: `routes.ts` conflicts (at least `tests/flows.spec.ts` + `client/client.js`), and the new failure-rollback in the install/update routes (`readManifestDeps`/`restoreManifestDeps`, #65) overlaps conceptually with `restoreBackup()`'s own retry/rollback loop — worth unifying rather than duplicating.

Happy to re-review quickly once these are addressed — the feature itself (#55) is very much wanted.

mintonight:
Thanks for the thorough review — all three blockers are addressed. The branch is rebased onto `main` @ `53c0956` (0 behind) and the head is now `39f7813`.

## 1. Test suite green on the branch

The auto-backup persist effect no longer crashes the client-test environment. We guard the storage access (`try/catch` around `localStorage.setItem`) instead of stubbing it, and additionally treat `localStorage` as optional everywhere (the credential fields are never written there at all anymore — see 2).

Full validation on the rebased head:
- `npm run typecheck` — clean (both tsconfigs)
- `npm run build` — clean (bundle regenerated against the rebased sources; the `client/client.js` conflict was resolved by rebuilding, not hand-merging)
- `npm test` — **134 tests / 14 files, all passing** (previously-red `tests/client/market-section.client.spec.tsx` included)
- `npm run test:compat` — 9 tests passing
- `node scripts/preflight.mjs` — ok

## 2. Security

- **WebDAV password** is never persisted in the browser anymore. [`savedWebdav()`](https://github.com/mintonight/dsh-market/blob/codex/backup-restore/src/client/MarketSection.tsx) resets `password: ''` on load; the fields in `localStorage` are URL/username only, and the password is re-entered per session as the UI note states.
- **SSRF hardening** is in `src/backup.ts`: WebDAV requests are https-only, refuse loopback/RFC1918/link-local IPv4 and `.internal`/`localhost` hostnames (incl. the cloud-metadata surfaces), `redirect: 'error'`, 30s timeout. `action:'restore'` no longer echoes the fetched body blindly — `downloadWebdav` runs `validatedBackup()` server-side and only a strict `ProfileBackup` is returned to the client; the real restore happens at `POST /dsh-market/restore`, which parses/validates again before writing.
- **`GET /dsh-market/backup`** is now gated by the same `trustedRestartRequest` (loopback + same-origin, no forwarding headers) as `/dsh-market/restart`. The backup format is unchanged (1:1 for faithful restores) but both the Local and WebDAV cards now carry a loud UI warning that profile config may hold credentials — including that WebDAV uploads should only go to trusted targets.

## 3. Rebase

Done — rebased onto current `main` (`53c0956`). Conflicts touched `tests/flows.spec.ts`, `src/routes.ts`, `src/client/MarketSection.tsx` and the bundle; all resolved (source merged, bundle regenerated). Regarding #65's rollback helpers: the branch keeps its own `restoreProfileBackup` rollback because it restores *profile state generally* (files), whereas #65 handles manifest/node_modules rollbacks for the install/update path — they operate on different layers and coexist without a shared abstraction, but I'm happy to unify in a follow-up if you'd prefer.

The restore-reinstall flow also gained a per-plugin fallback (bulk install failure → retry each plugin individually, surfacing failures without aborting the whole restore) — covered in `tests/flows.spec.ts`.

fkysly (approved):
Re-reviewed the rebased head — all three blockers are properly addressed, and the security work goes beyond what was asked:

- **Tests**: 151/151 on the branch merged with current main (the previously-red client suite included), typecheck + build + restart-smoke clean.
- **Credentials**: password is never persisted (`savedWebdav` resets it on load, only URL/username stored), with the same-origin rationale documented in code.
- **SSRF**: https-only, allowlist-based IPv6 policy (2000::/3 — which also kills the IPv4-mapped bypass), private/loopback/link-local IPv4 refused, and — nicely — the connection is pinned to the vetted IP with SNI preserved, closing the DNS-rebinding window between check and connect. `downloadWebdav` validates server-side and returns only a strict `ProfileBackup`, never a generic echo.
- **`GET /backup`** gated by `trustedRestartRequest` + loud bilingual credential warnings in both cards.

I pushed one commit onto the branch (`67cc6da`): the per-plugin restore fallback still read/wrote the manifest and probed node_modules through the ambient `profileDir(config.profile)` — in DSH Desktop (#72, merged after your rebase) the profile directory is host-authoritative and the ambient path points at the wrong profile. Everything else already used `activeProfileDir` correctly; these were the last two call sites.

On unifying `restoreProfileBackup`'s rollback with #65's manifest rollback: agreed they're different layers — a follow-up is fine, no blocker. Thanks for the thorough turnaround, @mintonight — this closes #55's core ask.

PR #67, opened 2026-08-15
fix(dshmarket): accept version-qualified names in approve-builds and allow retry of failed installs

ZAshenOne:
pnpm 11's ndjson `pnpm:ignored-scripts` event reports **version-qualified** package names (`cloudflared@0.7.3`), but the approve-builds route's `PKG_RE` and `existsSync(node_modules/<name>)` only accept bare package names. Every package was filtered out → `400 "no installed packages given"`, so the one-click **"Allow build scripts and retry"** flow silently did nothing.

Additionally, a failed `pnpm add` (blocked build scripts) already writes the dependency into package.json (manifest is written before the build-script check runs). On retry, the install route's duplicate guard treated that leftover as "already installed" ("已以「…」安装过同一个插件"), dead-ending the flow even after a successful approval.

## Changes
- **ndjson.ts**: strip `@version` suffixes from `pnpm:ignored-scripts` names, matching the human-line fallback in `install.ts`'s `parseIgnoredBuilds`.
- **routes.ts (approve-builds)**: defensively strip `@version` suffixes before `PKG_RE`/`existsSync` validation, so version-qualified names from pnpm 11 ndjson are accepted.
- **routes.ts (install)**: only reject the duplicate guard when the matched dependency is **active** (bundle layer or live mount) or a genuinely different source; a same-source leftover of a failed install is retried and excluded from the `before` set so post-install validation and hot-mount treat it as newly added.
- **routes.ts (install)**: surface a friendly bilingual message instead of pnpm's raw stack when build scripts are blocked.
- **tests/ndjson.spec.ts**: regression test for version-stripping (incl. scoped names + dedupe).

## Repro
```sh
dsh plugin --profile web add @linxin666/dsh-web-ui-all  # transitive deps cloudflared/cpu-features/ssh2 have build scripts
```
Expected: `ERR_PNPM_IGNORED_BUILDS` exit 1 with the dep already written to package.json; clicking "Allow build scripts and retry" currently 400s and dead-ends. After this fix the approval is accepted and the retry install completes (exit 0, bundle activated).

## Test status
- `tsc -p tsconfig.json --noEmit`: clean
- `vitest run`: 93/93 pass (12 files), including the new regression test

### DIFF
--- a/src/ndjson.ts
+++ b/src/ndjson.ts
@@ -138,7 +138,13 @@ export function createProgressTracker(): ProgressTracker {
       snap.seen = true
       if (Array.isArray(msg.packageNames)) {
         for (const pkg of msg.packageNames) {
-          if (typeof pkg === 'string' && !snap.ignoredBuilds.includes(pkg)) snap.ignoredBuilds.push(pkg)
+          // pnpm's ndjson event reports version-qualified names (cloudflared@0.7.3);
+          // the approve-builds allowlist keys and node_modules lookups use bare
+          // package names, so strip the suffix (same rule as the human-line
+          // fallback in install.ts's parseIgnoredBuilds).
+          const at = typeof pkg === 'string' ? pkg.lastIndexOf('@') : -1
+          const bare = at > 0 ? pkg.slice(0, at) : pkg
+          if (typeof bare === 'string' && bare !== '' && !snap.ignoredBuilds.includes(bare)) snap.ignoredBuilds.push(bare)
         }
       }
       return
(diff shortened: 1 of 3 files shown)


fkysly (approved):
Verified locally with main (post #70/#74) merged in: 108/108 tests, typecheck clean, no conflicts. The version-strip matches the human-line fallback rule, and the retry path correctly requires same-source AND not-active (bundle/live) before bypassing the duplicate guard — with unreadable-manifest defaulting to active/blocked, which is the safe direction.

Note for the record: this fixes the approve-builds 400 and the retry dead-end, which is a prerequisite for — but not all of — #68 (git-hosted `prepare` rejected before materialization needs a different allowBuilds key) and #69 (the update route still has no ignored-builds handling). Follow-ups for those are in the works. Thanks @ZAshenOne!

PR #70, opened 2026-08-15
fix: live check must match bundle entries named as package subpaths

### LINKED ISSUE #71: Installed plugins whose bundle patch names a subpath entry always show "restart" — even when running
## What happened / 发生了什么

Installed bundle-layer plugins whose `cordis.patch.yml` names a **subpath entry** (rather than the bare package name) permanently show **"重启后生效 / restart required"** in the plugin inventory — **even after a restart, even though the plugin is running** (its loader fiber is up and its tools/skills are available).

Expected: a bundle-layer plugin whose fiber is up should read as **live / 已热加载** in Settings → Plugin Market.

Observed cases:
- `@vectorize-io/hindsight-coding-agents` → patch row `- id: hindsight, name: "@vectorize-io/hindsight-coding-agents/dsh"` → shows "restart"
- `aegis` → patch row `- id: aegis-method-pack, name: 'aegis/extensions/dsh/index.js'` → shows "restart"

While `dsh-plugin-knowledge-graph` and `@linxin666/dsh-web-ui-all` (patch rows naming the bare package) correctly show **live**.

## Root cause / 根因

`liveNames()` in `src/routes.ts` builds the live set from **loader entry names** — the `name:` field of each bundle patch row. `verifyActivation()` in `src/verify.ts` then looks the **package name** up with `live.has(name)`. When the patch names a subpath (`@scope/pkg/dsh` or `pkg/extensions/dsh/index.js`), the exact-name lookup never matches, so the state falls through to `restart` — even though the entry's fiber is up.

## Steps to reproduce / 复现步骤

1. Install a plugin whose bundle patch names a subpath entry, e.g. `@vectorize-io/hindsight-coding-agents` (patch: `name: "@vectorize-io/hindsight-coding-agents/dsh"`) or `aegis` (patch: `name: 'aegis/extensions/dsh/index.js'`).
2. Restart DSH so the bundle layer loads the plugin.
3. Open Settings → Plugin Market → Installed. The plugin shows "已进入 profile bundle 层但本次未能热挂载;重启后生效 / in the bundle layer but not hot-mounted this session — it activates on restart", even though it is actually running (its tools/skills are live in sessions).

## Environment / 环境

- OS: Windows 11 x64
- dsh: DeepSeek Harness desktop (Web GUI, profile `web`)
- dshmarket: 1.5.0

## Evidence / 证据

`GET /dsh-market/installed` on a running host, after a restart that loaded both plugins:

```json
"@vectorize-io/hindsight-coding-agents": { "state": "restart", "bundle": true, "hot": false },
"aegis": { "state": "restart", "bundle": true, "hot": false },
"dsh-plugin-knowledge-graph": { "state": "live", "bundle": true, "hot": true }
```

At the same time the Hindsight memory tools (`hindsight_*`) and the Aegis skill pack were live in the session — confirming the plugins were loaded, and only the inventory label was wrong.

## Fix / 修复

PR #70 — match the package name itself or any `name/` subpath entry (with a `/` bound so similarly-prefixed different packages never match). Regression tests included.

wtiaw:
## Problem / 问题

`verifyActivation` reports a running bundle-layer plugin as **"restart"** ("已进入 profile bundle 层但本次未能热挂载;重启后生效") in the plugin inventory after every boot, even though its loader fiber is up — the plugin is actually working.

`liveNames()` (routes.ts) builds the live set from **loader entry names** — the `name:` field of each bundle patch row. Most bundles name the bare package (`dshmarket`, `@scope/pkg`), which matches the package-name lookup in `verifyActivation` exactly. But bundles that point at a **subpath entry** never match:

- `@vectorize-io/hindsight-coding-agents` → patch row `name: "@vectorize-io/hindsight-coding-agents/dsh"`
- `aegis` → patch row `name: "aegis/extensions/dsh/index.js"`

For these, `live.has(packageName)` is always false, so the inventory shows "restart" indefinitely — misleading users into restarting (or believing the plugin is broken) when it is already live.

## Fix / 修复

Match the package name itself **or any `name/` subpath entry** in the live set, with the `/` bound so a similarly-prefixed different package (e.g. `@scope/pkg2` vs `@scope/pkg`) never reads as live.

## Tests / 测试

Adds regression tests (red first) covering:
- scoped subpath entry: `@vectorize-io/hindsight-coding-agents/dsh` → live
- unscoped subpath entry: `aegis/extensions/dsh/index.js` → live
- prefix-bound: `dsh-loop-tool` must NOT make `dsh-loop` live

## Verification / 验证

- `npm test`: 95 passed (12 files)
- `npm run typecheck`: clean

Fixes #71 — bundle-layer subpath entries always read as "restart".

### DIFF
--- a/src/verify.ts
+++ b/src/verify.ts
@@ -53,6 +53,24 @@ interface PkgDsh {
   client?: unknown
 }
 
+/**
+ * True when `live` contains the package itself or a subpath entry of it.
+ *
+ * The live set (see `liveNames` in routes.ts) holds loader entry names — the
+ * `name:` field of each bundle patch row. Bundles usually name the bare
+ * package (`dshmarket`, `@scope/pkg`), but may point at a subpath entry
+ * (`@vectorize-io/hindsight-coding-agents/dsh`, `aegis/extensions/dsh/index.js`).
+ * Either form means the package's fiber is up and it must read as live;
+ * a different package sharing a name prefix (`@scope/pkg2` vs `@scope/pkg`)
+ * must not — the `/` bound keeps the match a real subpath.
+ */
+function liveIncludes(live: ReadonlySet<string>, packageName: string): boolean {
+  if (live.has(packageName)) return true
+  const prefix = `${packageName}/`
+  for (const name of live) if (name.startsWith(prefix)) return true
+  return false
+}
+
 function readPkgDsh(profile: string, name: string): PkgDsh | null {
   try {
     const manifest = JSON.parse(
@@ -110,7 +128,7 @@ export function verifyActivation(
     }
   }
 
-  if (live.has(name)) {
+  if (liveIncludes(live, name)) {
     const clientOnly = dsh.bundle === undefined && dsh.client !== undefined
     return {
       state: 'live',
--- a/tests/verify.spec.ts
+++ b/tests/verify.spec.ts
@@ -92,4 +92,32 @@ describe('verifyActivation (P0-2)', () => {
     pkg('dsh-loop', { dsh: { bundle: { patch: './cordis.patch.yml' } }, main: 'index.js' }, { 'index.js': '', 'cordis.patch.yml': SIMPLE_PATCH })
     expect(verifyActivation('web', 'dsh-loop', new Set())).toMatchObject({ state: 'restart', bundle: true })
   })
+
+  it('live when the bundle entry name is a scoped subpath of the package (patch name ≠ package name)', () => {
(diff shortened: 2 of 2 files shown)


fkysly (approved):
Verified locally on top of main @ 7686775: the two new subpath tests fail against main's verify.ts and pass with the fix; full suite 95/95, typecheck clean. The `/`-bound prefix match is exactly right — `dsh-loop-tool` can no longer read as `dsh-loop` live. Thanks for the red-first tests, @wtiaw!

Issue #68, opened 2026-08-15
[Bug] git-hosted plugins with a prepare script fail with ERR_PNPM_GIT_DEP_PREPARE_NOT_ALLOWED and the approve-builds flow can never approve them


yzr278892:
### What happened / 发生了什么

Installing `dsh-github-intelligence` (registry source `https://github.com/zoahdev/dsh-github-intelligence`, `npm: null`, install target `github:zoahdev/dsh-github-intelligence`) from the market fails. The UI shows a truncated raw error ending in `" allowlist.` followed by the pnpm stack (`preparePackage` → `prepareGitHostedPkg` → `fetch`).

Full pnpm error (reproduced with pnpm 11.21.0 and the exact web-profile workspace settings):

```
[ERR_PNPM_GIT_DEP_PREPARE_NOT_ALLOWED] Failed to prepare git-hosted package fetched from "https://codeload.github.com/zoahdev/dsh-github-intelligence/tar.gz/<commit>": The git-hosted package "dsh-github-intelligence@2.8.0" needs to execute build scripts but is not in the "allowBuilds" allowlist.
```

### Why / 原因

The plugin's `package.json` contains:

```json
"scripts": {
  "prepare": "pnpm run build"
}
```

pnpm 10+/11 requires git-hosted dependencies to be explicitly allowed in `allowBuilds` before running prepare/prepack scripts. This rejection happens inside pnpm's fetcher **before the package is materialized in `node_modules` and before `package.json` is written**.

The market's one-click "Allow build scripts and retry" (`POST /dsh-market/approve-builds`) filters candidates with:

```ts
existsSync(join(profileDir(config.profile), 'node_modules', name, 'package.json'))
```

For this failure mode that path never exists, so the route can never approve the package. `classifyPnpmFailure()` also does not recognize `ERR_PNPM_GIT_DEP_PREPARE_NOT_ALLOWED`, so the raw pnpm wall-of-text is surfaced and the actionable hint is lost/truncated.

### Steps to reproduce / 复现步骤

1. In the market, install `dsh-github-intelligence` (or run `dsh plugin --profile web add github:zoahdev/dsh-github-intelligence`).
2. The profile's `pnpm-workspace.yaml` does not already contain this git package in `allowBuilds`.
3. Observe the error above.
4. After the failure the plugin is in neither `package.json` nor `node_modules`, and the approve-builds flow cannot fix it.

### Manual workaround / 临时绕过

[...]

PR #72, opened 2026-08-15
feat: use DSH Desktop profile and pnpm services

labmimors:
## What changed

- Detect DSH Desktop through the public generation-scoped `desktopProfiles` service.
- Keep the Desktop-only `desktopPnpm` dependency inside a nested `ctx.inject()` callback; ordinary DSH still uses the existing config/argv/CLI path.
- Use `desktopProfiles.current.name` and `.dir` as the authoritative profile identity and filesystem root, including Unicode/spaced profile names.
- Adapt `desktopPnpm.runPlugin()` streams, cancellation, timeout, teardown, and generation-wide busy errors to the existing market operation surface.
- Route status, setup, cancel, install, update, rollback, build approval, source/name guards, themes, and activation verification through the injected operation runtime and host-authoritative profile directory.
- Preserve a dependency-free boundary: there is no runtime import from or hard dependency on `dsh-plugin-desktop`.

## Rebase safety

This branch is rebased onto `main` at `eb7881f` (the 1.5.1 release, including #76/#77/#78). The newly added profile-sensitive paths all receive the same `activeProfileDir`:

- `setAllowBuilds` plus installed-spec lookup for `gitAllowBuildsKey`;
- `readManifestDeps` / `restoreManifestDeps` for failed install and update rollback;
- duplicate, inactive-leftover retry, and same-name/different-source guards;
- installed versions, lockfile commits, activation checks, and hot mounts.

The flow suite now uses a Desktop-owned directory whose profile name is intentionally not derivable from `$DSH_HOME`, and verifies build approval/git keys, failed-install rollback, failed-update rollback, and the same-name source guard against that directory while the ambient `web` profile remains unchanged.

## Desktop contract evidence

DSH Desktop's supported plugin contract explicitly guarantees that `desktopProfiles` is registered before Loader entries and prescribes the presence check plus nested `desktopPnpm` injection used here:

- [Supported cross-environment contract (pinned source)](https://github.com/anywhere-labs/deepseek-harness-desktop/blob/4f68147091e585aaa1d815f99d30a657b3842d7c/dsh-plugin-desktop/docs/plugin-services.md#L190-L243)
- [Launcher registration before Host boot proceeds (pinned source)](https://github.com/anywhere-labs/deepseek-harness-desktop/blob/4f68147091e585aaa1d815f99d30a657b3842d7c/dsh-plugin-desktop/src/main.ts#L159-L190)

## Lifecycle and failure behavior

- Desktop profile identity never falls back to `web` after `desktopProfiles` is present.
- Package work is cancelled and awaited when the owning Cordis effect is disposed.
- Timeout sends both the public `AbortSignal` and `handle.cancel()` process-tree cancellation paths; a regression test covers a provider that ignores the signal and settles only after `cancel()`.
- A Desktop generation-wide operation gate is returned as HTTP `409`, rather than flattened into a generic install failure.
- The existing self-restart endpoint remains available in ordinary DSH and is disabled in Desktop mode because restart is not part of the public plugin contract.

## Validation

- `npm ci` — lockfile install completed; audit reported 0 vulnerabilities
- `npm run check` — both TypeScript compiler faces, production build, and restart smoke passed
- `npm test` — 15 files, 141 tests passed
- `npm run test:compat` — 9 pnpm 9/10/11 compatibility tests passed
- `npm run test:web` — 5 tests discovered and skipped because this isolated checkout has no reachable DSH CLI
- `node scripts/preflight.mjs` / `validate-registry.mjs` / `smoke-spawn.mjs` — all passed (registry retained its 12 allowed display-name-collision warnings)
- `npm pack --ignore-scripts --dry-run --json` — 64 entries; package contains `LICENSE` and `lib/index.js`
- `npm audit --omit=dev --audit-level=high` — 0 vulnerabilities
- Independent final review: PASS after the timeout cancellation regression was fixed

## Real Desktop manual gate — not claimed

No installed or runnable DSH Desktop build was found on this machine, and this change did not install a system application. Therefore I have **not** performed or claimed a real Electron install + cancel + generation-busy `409` run. That remains an explicit maintainer/release-side manual gate before shipping a release; the fake-service and route tests above do not substitute for it.

GitHub has also classified the current fork CI run as [`action_required`](https://github.com/dsh-market/dsh-market/actions/runs/31919615441) without starting any jobs. A maintainer must approve that fork workflow before its Ubuntu, Windows, compatibility, and real-DSH web lanes can report.

## Disclosure

I maintain `@labmimors/dsh-mcp-lens`, a DSH plugin that would benefit from a Desktop-compatible marketplace. This patch is intentionally marketplace infrastructure for every curated plugin; it does not add or special-case MCP Lens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,23 @@ dsh plugin --profile web add dshmarket
 
 Restart `dsh web`, then open **Settings → Plugin Market**.
 
+### DSH Desktop
+
+On DSH Desktop builds that expose the public `desktopProfiles` and
+`desktopPnpm` services, install from **Open DSH Terminal** with:
+
+```sh
+dsh plugin add dshmarket
+```
+
+The market automatically targets the immutable active Desktop profile and
+uses Desktop's packaged pnpm operation service. It does not probe or install a
+system pnpm, and Desktop keeps ownership of application restart. Ordinary DSH
+continues to use the existing profile/argv/CLI path when those services are
+absent. The detection order follows Desktop's supported
+[cross-environment plugin contract](https://github.com/anywhere-labs/deepseek-harness-desktop/blob/4f68147091e585aaa1d815f99d30a657b3842d7c/dsh-plugin-desktop/docs/plugin-services.md#cross-environment-plugin-optional-desktop-adapter-and-ordinary-dsh-fallback).
+This compatibility does not mean the market is bundled with Desktop.
+
 ## What you get
 
 - **Browse & search** the full community catalog (300+ plugins, growing daily) — category filters, star counts, top/new sorting, bilingual descriptions that follow your UI language
@@ -68,7 +85,7 @@ Live from [awesome-dsh-plugin.com/plugins.json](https://awesome-dsh-plugin.com/p
 
 ### DeepSeek Harness Desktop
 
-[DeepSeek Harness Desktop](https://github.com/anywhere-labs/deepseek-harness-desktop) is a modern desktop client for the DeepSeek Harness ecosystem — start and manage a local Harness service without installing Node.js or touching the command line. Plugin marketplace, mobile remote control, and IM Channels are on its roadmap.
+[DeepSeek Harness Desktop](https://github.com/anywhere-labs/deepseek-harness-desktop) is a modern desktop client for the DeepSeek Harness ecosystem — start and manage a local Harness service without a system Node.js installation. Compatible builds can host a user-installed dsh-market through their public profile and package-operation services; bundling remains a Desktop project decision.
 
 [Website](https://www.dshdesktop.cn) · [GitHub](https://github.com/anywhere-labs/deepseek-harness-desktop)
 
--- a/README.zh.md
+++ b/README.zh.md
@@ -25,6 +25,21 @@ dsh plugin --profile web add dshmarket
 
 重启 `dsh web`，打开 **设置 → 插件市场**。
 
+### DSH Desktop
+
+如果 DSH Desktop 版本提供公开的 `desktopProfiles` 与 `desktopPnpm`
+服务，可从 **Open DSH Terminal** 安装：
+
+```sh
+dsh plugin add dshmarket
+```
+
+市场会自动使用该代 Desktop 已锁定的活动 profile，以及 Desktop 内置的
+pnpm 操作服务；不会探测或安装系统 pnpm，应用重启仍由 Desktop 负责。
+没有这些服务的普通 DSH 会继续使用现有 profile/argv/CLI 路径。检测顺序遵循
+Desktop 支持的[跨环境插件 contract](https://github.com/anywhere-labs/deepseek-harness-desktop/blob/4f68147091e585aaa1d815f99d30a657b3842d7c/dsh-plugin-desktop/docs/plugin-services.zh.md#跨环境插件可选-desktop-adapter-与普通-dsh-fallback)。
+兼容 Desktop 并不代表市场已预装或捆绑进 Desktop。
+
 ## 你会得到
 
 - **逛与搜**——完整社区目录（300+ 插件，每天在涨），分类筛选、star 数、最热/最新排序，中英描述跟随界面语言
@@ -68,7 +83,7 @@ dsh plugin --profile web add dshmarket
 
 ### DeepSeek Harness Desktop
 
-[DeepSeek Harness Desktop](https://github.com/anywhere-labs/deepseek-harness-desktop) 是一款为 DeepSeek Harness 生态打造的现代化桌面端，让用户无需配置 Node.js 或执行命令，即可启动和管理本地 Harness 服务。项目后续还将支持插件市场、移动端远程控制和 IM Channels。
+[DeepSeek Harness Desktop](https://github.com/anywhere-labs/deepseek-harness-desktop) 是一款为 DeepSeek Harness 生态打造的现代化桌面端，无需在系统中安装 Node.js 即可启动和管理本地 Harness 服务。兼容版本可通过公开的 profile 与包操作服务运行用户自行安装的 dsh-market；是否预装仍由 Desktop 项目决定。
 
 [访问官网](https://www.dshdesktop.cn) · [GitHub](https://github.com/anywhere-labs/deepseek-harness-desktop)
 
--- a/src/dsh-cli.ts
+++ b/src/dsh-cli.ts
@@ -10,7 +10,7 @@
 import { spawn } from 'node:child_process'
 import type { ChildProcess } from 'node:child_process'
 import { homedir } from 'node:os'
-import { dirname, join, resolve } from 'node:path'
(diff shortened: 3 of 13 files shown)


fkysly (commented):
Reviewed in depth (locally: this branch's own suite is green, 101/101; typecheck clean). The architecture is the right shape — an optional `PluginCommandRuntime` boundary consumed by the routes, an explicit `profileDirectory` threaded through every profile reader with the DSH_HOME derivation as default, structural typing of `desktopProfiles`/`desktopPnpm` with no hard dependency on the Desktop package, and the operation-gate → 409 mapping. The relaxed `PROFILE_RE` only when the directory is host-authoritative is a nice touch.

Three things before merge:

1. **The rebase onto today's main is NOT mechanical.** #76/#77/#78 added new profile-directory-dependent calls in the very routes this PR rewires: `setAllowBuilds` + `gitAllowBuildsKey` (approve-builds, #68/#69), `readManifestDeps`/`restoreManifestDeps` (failure rollback, #65), and the name-clash guard (#66). All of them derive the directory from `profileDir(config.profile)` — after rebase they must accept/receive `activeProfileDir` too, or the Desktop path will read/write the WRONG directory for allowBuilds and rollback while looking perfectly green in ordinary-DSH tests. Please extend them in the same pass.

2. **Service-timing assumption**: `ctx.get('desktopProfiles')` at apply time decides the mode permanently. If Desktop ever registers the service after the market's `apply` runs, the market silently falls into the ordinary-CLI path inside Desktop. The comment says Desktop registers it before Loader entries — can you point at the Desktop-side guarantee (doc or code), so this doesn't rot?

3. **Real-Desktop verification**: the `desktopPnpm` contract here is exercised only against the in-test fake. Before this ships in a release, someone needs one manual run against an actual DSH Desktop build (install + cancel + generation-busy 409). If you've done that, please note the Desktop version in the PR body.

Marking as maintainer-review-passed on architecture; holding merge until the rebase lands (happy to help with the conflict resolution given the overlap is with changes I merged today).

labmimors:
Thanks for the detailed review. I addressed each gate on the final rebased head `a98347f`:

1. **Today's `main` / #76–#78:** rebased onto the later `eb7881f` 1.5.1 release commit. `setAllowBuilds`, the installed-spec lookup feeding `gitAllowBuildsKey`, `readManifestDeps` / `restoreManifestDeps`, and the duplicate/inactive-leftover/name-clash guards now all read or write through `activeProfileDir`. The same propagation covers the update version/lock checks and activation paths. New Desktop-directory flow tests fail if any of those operations fall back to ambient `profileDir(config.profile)`.

2. **Service timing:** the Desktop-side supported contract explicitly says the launcher registers `desktopProfiles` before Loader entries and prescribes exactly this presence-check + nested-`desktopPnpm` pattern: https://github.com/anywhere-labs/deepseek-harness-desktop/blob/4f68147091e585aaa1d815f99d30a657b3842d7c/dsh-plugin-desktop/docs/plugin-services.md#L190-L243 . The launcher source registers `DesktopProfileService` in its Host boot callback here: https://github.com/anywhere-labs/deepseek-harness-desktop/blob/4f68147091e585aaa1d815f99d30a657b3842d7c/dsh-plugin-desktop/src/main.ts#L159-L190 . Both links are now recorded in the code/README so the assumption is reviewable rather than implicit.

3. **Real Desktop verification:** not done and not claimed. Read-only local inventory found no installed/runnable DSH Desktop build, and I did not install a system app. A real Electron install + cancel + generation-busy `409` run remains the explicit maintainer/release-side manual gate.

Current automated evidence: `npm run check` passed; 15 test files / 141 tests passed; pnpm 9/10/11 compatibility 9/9 passed; preflight, registry validation, and spawn smoke passed; package dry-run includes `LICENSE` and `lib/index.js`; production audit reports 0 vulnerabilities. Web e2e discovered 5 tests but skipped them because this checkout has no reachable DSH CLI. An independent final review returned PASS after one timeout-lifecycle issue was fixed by invoking both `AbortSignal` and `handle.cancel()` and adding a regression test.

The current GitHub Actions run is waiting for maintainer approval (`action_required`, with no jobs started): https://github.com/dsh-market/dsh-market/actions/runs/31919615441 . Once approved, it will supply the upstream Ubuntu/Windows, compatibility, and real-DSH web lanes that this contributor account cannot start itself.

fkysly (approved):
Re-verified the rebased head locally: 141/141 tests, typecheck clean, merges clean with main (now on top of the 1.5.1 release commit). Every gate from the previous review is addressed:

1. The #76–#78 surfaces are all correctly threaded — `setAllowBuilds(…, activeProfileDir)`, the `specs` feeding `gitAllowBuildsKey` read via `readInstalled(profile, activeProfileDir)`, both `readManifestDeps`/`restoreManifestDeps` rollback sites, the node_modules presence checks, and the name-clash guard. The no-ambient-fallback tests are exactly the right insurance.
2. Service-timing assumption is now anchored to the documented Desktop contract and launcher source — reviewable, not implicit. Thanks for the links.
3. Real-Desktop verification honestly deferred — agreed, that stays a **release-side manual gate**: this merges now, but before the next release that includes it, someone runs install + cancel + generation-busy 409 against an actual Desktop build. The Desktop path is inert unless `desktopProfiles` exists, so ordinary DSH users are unaffected in the meantime.

Excellent turnaround, @labmimors.

fkysly:
**Real-Desktop verification complete** — the release-side manual gate is cleared. ✅

Environment: DSH Desktop **2.0.0** (arm64, notarized build from `anywhere-labs/deepseek-harness-desktop` v2.0.0), macOS (darwin 25.3.0), packaged pnpm **11.7.0**, dshmarket built from main @ `54f1f0d` (this PR's merge) installed into the `desktop` profile as a `file:` tarball + bundle entry.

| Check | Result |
|---|---|
| Profile identity | ✅ `/dsh-market/installed` reports `profile: "desktop"` — no fallback to `web` |
| Install via `desktopPnpm` | ✅ curated npm plugin installed through the packaged pnpm; **ndjson structured progress works through the Desktop stream adapter** (`ndjson: true`, staged phases); activation inventory correct (`restart` for a config-row bundle patch) |
| Cancel mid-install | ✅ `/dsh-market/cancel` → install response `cancelled: true, partial: false, changed: []` — clean abort, zero manifest residue (the AbortSignal + `handle.cancel()` path) |
| Concurrent operation | ✅ second install during an active one → **HTTP 409** `another install is already running` |
| Self-restart disabled | ✅ `POST /dsh-market/restart` → **HTTP 403** `self-restart is disabled for this host` |
| Uninstall | ✅ both test plugins removed cleanly |

One honest caveat: the Desktop **generation-wide** busy gate (`desktopPnpm` throwing synchronously because Desktop itself holds an operation) is not reachable via the market's HTTP surface alone — the market's own `installing` flag always intercepts market-originated concurrency first. That specific mapping remains covered by `tests/desktop-runtime.spec.ts` only. Everything else ran against the real contract.

#72 is clear to ship in the next release.

PR #80, opened 2026-08-16
fix(cli): avoid Node DEP0190 shell+argv spawn on Windows

tatakaria:
## Problem

Node 24 emits `DEP0190 DeprecationWarning` on every market status probe
(`pnpm --version`) and on install/update runs on Windows, because
`spawn(file, args, { shell: true })` passes argv alongside a shell — Node
concatenates the arguments unescaped, which is deprecated.

Windows `pnpm`/`npm`/`dsh` are `.cmd` shims that still require a shell, so
the flag cannot simply be removed.

## Change

- Add `spawnShim()`: the Windows shim path builds an explicitly quoted
  command line (`quoteCmdArg` / `cmdCommandLine`) and spawns
  `cmd.exe /d /s /c` with `shell: false` + `windowsVerbatimArguments`.
- Migrate `probePnpm`, `runQuiet`, and `runDshPlugin` to `spawnShim`.
  POSIX behavior is unchanged (direct, shell-free spawn).
- Update `scripts/smoke-spawn.mjs` to exercise the same shim path in CI.
- Add unit tests for cmd.exe quoting and command-line assembly.

## Verification

- `npm run typecheck`
- `npm test` — 131 passed
- `npm run build`
- `node scripts/preflight.mjs`
- `node scripts/restart-smoke.mjs`
- `node scripts/smoke-spawn.mjs` — `smoke ok on win32: npm + pnpm spawn via cmd.exe shim`
- `probePnpm()` on Windows returns `true` with no DEP0190 warning

### DIFF
--- a/scripts/smoke-spawn.mjs
+++ b/scripts/smoke-spawn.mjs
@@ -2,16 +2,29 @@
 /**
  * Cross-platform smoke for the tool-spawning pattern used by routes.ts.
  * On Windows, npm/corepack/pnpm are `.cmd` shims that node:child_process
- * cannot start without a shell (#2/#3/#5) — this guards the `shell` flag
- * from regressing. Runs in CI on windows-latest and ubuntu-latest.
+ * cannot start without a shell (#2/#3/#5) — this guards the cmd.exe shim
+ * path from regressing. Runs in CI on windows-latest and ubuntu-latest.
  */
 import { spawn } from 'node:child_process'
 
-const winCmdShim = process.platform === 'win32'
+const CMD_METACHARS = /[\s"&|<>^()%!]/
+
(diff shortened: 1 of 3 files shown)


fkysly (approved):
Verified locally merged with main: 145/145 tests, typecheck clean. The `cmd /d /s /c` + outer-quote + `windowsVerbatimArguments` recipe is the established correct pattern for .cmd shims, and the quoting is strictly better than the deprecated shell+argv concat it replaces. POSIX is provably unchanged (`viaShell` is only ever `winCmdShim`), and CI's Windows job runs the updated `smoke-spawn.mjs` against the real shim path. Thanks @tatakaria!

PR #115, opened 2026-08-16
fix(profile): quote reserved YAML keys in the allowBuilds block

Bruce-Yii:
## Summary

`setAllowBuilds` in `src/profile.ts` writes package keys into the profile's `pnpm-workspace.yaml` `allowBuilds` block **verbatim**. Scoped npm names start with `@` — a reserved YAML indicator that cannot begin a plain scalar — so approving a build for e.g. `@deepseek-ai/dsh-subprocess-local` produces:

```yaml
allowBuilds:
  @deepseek-ai/dsh-subprocess-local: true
```

which is invalid YAML. Every later pnpm run in that profile (install, update, even the market's own flows) then fails with a misleading `bad indentation of a mapping entry` error — the file is bricked until manually repaired.

## Change

- **Emit side**: quote a key whenever a plain scalar cannot represent it — reserved leading indicators (`@ % ` " '` and friends), or `: ` / trailing `:`. Git keys like `name@git+https://github.com/owner/repo.git` (#68) keep their existing plain form.
- **Parse side**: when re-reading the block, strip surrounding single/double quotes from captured keys so a later rewrite never nests quotes (`''@scope/pkg''`).

## Tests

Two new cases in `tests/profile.spec.ts`:
- scoped `@` keys are written quoted and stay valid YAML,
- an already-quoted scoped key round-trips without quote nesting.

Verified locally: `vitest run tests/profile.spec.ts` — 13 tests passed (11 existing + 2 new); `tsc -p tsconfig.json --noEmit` clean.

This is the bug we hit on a real machine: the market approved a build for `@deepseek-ai/dsh-subprocess-local` while installing `dsh-browser`, and the resulting unquoted key broke `pnpm-workspace.yaml` for every subsequent operation.

### DIFF
--- a/src/profile.ts
+++ b/src/profile.ts
@@ -242,6 +242,28 @@ export function pluginSubdirs(root: string): string[] {
  * (#6 by @qichuang321.)
  * @returns every package now allowed.
  */
+/**
+ * Quote a YAML block-mapping key when a plain scalar would be invalid.
+ * Scoped npm names start with `@` — a reserved YAML indicator — so an
+ * unquoted `@scope/pkg: true` entry breaks the whole pnpm-workspace.yaml
+ * for every later pnpm run in the profile (and for the market itself).
+ * Keys containing `: ` or ending with `:` are quoted for the same reason;
+ * git keys like `name@git+https://…` keep their existing plain form.
+ */
+function quoteYamlKey(key: string): string {
+  if (/^[-?:,[\]{}#&*!|>'"%@`]/.test(key) || /:(\s|$)/.test(key)) {
+    return `'${key.replace(/'/g, "''")}'`
+  }
+  return key
+}
+
+/**
+ * Allow the given packages' build scripts in the profile's
+ * pnpm-workspace.yaml `allowBuilds` block (the key dsh profiles use),
+ * merging with existing entries and leaving the rest of the yaml intact.
+ * (#6 by @qichuang321.)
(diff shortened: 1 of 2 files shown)


fkysly (approved):
Confirmed as a real, self-inflicted bug in the #68 code path, and verified both directions on real pnpm 11.22:

**Before** — an unquoted scoped key is not just cosmetically wrong, pnpm refuses the file outright:
```
allowBuilds:
  @deepseek-ai/dsh-subprocess-local: true
   ^ bad indentation of a mapping entry
```
So approving a build for any `@scope/pkg` bricked the profile for **every** later pnpm run — including the market's own — until hand-repaired. Given #82 showed `@deepseek-ai/dsh-subprocess-local` and `@google/genai` are exactly the packages users get asked to approve, this was firing in the wild.

**After** — your emitter produces valid YAML that pnpm 11.22 parses clean:
```yaml
allowBuilds:
  '@deepseek-ai/dsh-subprocess-local': true
  koffi: true
  x@git+https://github.com/o/r.git: true
```

The scoping is right: quote only when a plain scalar genuinely cannot represent the key (reserved leading indicator, `: `, trailing `:`), leave everything else — including the #68 git keys — untouched, and unquote symmetrically on the parse side so rewrites round-trip. 221/221 locally.

Thanks @Bruce-Yii — this one had teeth.

PR #117, opened 2026-08-16
fix(install): recognize pnpm fetch timeouts and retry with a longer fetchTimeout

Bruce-Yii:
## Summary

pnpm's per-request fetch timeout aborts large tarball downloads. A `github:` source fetches the **whole repository** even for a `#path:` subdirectory plugin, and on slow networks pnpm's default 60-second per-request limit is simply not enough — the download is aborted with a DOMException (`The operation was aborted due to timeout`, code 23), logged by pnpm as `GET … error (23)` with internal retries.

That abort shape was **not** recognized by `isTransientPnpmFailure`, so the market neither retried nor explained it — users saw a raw `TimeoutError` wall of text for a perfectly installable plugin. Worse, a plain retry (the transient-network recovery) would fail again at the same 60s limit.

## Change

- **`src/pnpm-compat.ts`**: classify the abort signature (`operation was aborted due to timeout`, `TimeoutError`, `error (23)`) as a new `fetch-timeout` failure with an actionable bilingual message.
- **`src/install.ts`**: `withHoistRecovery` retries `add`/`remove` once with `--config.fetchTimeout=600000` (one-shot override, scoped like `RELEASE_AGE_OVERRIDE`) — a 10-minute per-request budget that survives multi-hundred-MB monorepo tarballs on slow links.

## Tests

- `tests/pnpm-compat.spec.ts`: the exact pnpm/undici abort text classifies as `fetch-timeout` (and not `transient-network`); unrelated output stays unrecognized.
- `tests/install.spec.ts`: a fetch-timeout failure triggers exactly one retry carrying `FETCH_TIMEOUT_OVERRIDE`; an already-overridden argv is not retried again and the final failure message reaches the UI.

Local: 223/224 tests pass. The single failure (`tests/backup.spec.ts` symlink creation) also fails on a clean checkout on this machine — Windows requires elevation to create symlinks, unrelated to this change.

## Background

Encountered in the wild: installing `@openviking/dsh-memory-plugin` (`github:volcengine/OpenViking#path:/examples/dsh-memory-plugin`) — a ~100MB monorepo tarball on a network that reaches codeload at ~300KB/s — timed out repeatedly before the fix; with the longer fetchTimeout the same download completes.

### DIFF
--- a/src/install.ts
+++ b/src/install.ts
@@ -16,7 +16,16 @@ import { logEvent } from './log.ts'
 export const RELEASE_AGE_OVERRIDE = '--config.minimumReleaseAge=0'
 
 /**
- * Run one plugin command with automatic recovery from two known pnpm traps:
+ * Longer per-request fetch timeout for one retried command. pnpm's default
+ * 60-second limit aborts large tarball downloads (github: sources fetch the
+ * WHOLE repo even for a `#path:` subdirectory plugin) on slow networks; a
+ * plain retry fails again at the same limit, so the recovery re-runs with
+ * this override once. Scoped to a single command like RELEASE_AGE_OVERRIDE.
+ */
+export const FETCH_TIMEOUT_OVERRIDE = '--config.fetchTimeout=600000'
+
+/**
+ * Run one plugin command with automatic recovery from three known pnpm traps:
  *
  * - pnpm-major drift (#20 bug 2): a modules directory built by a different
  *   pnpm major fails mutation; pnpm's documented remedy is one `install` to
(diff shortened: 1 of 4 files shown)


fkysly (approved):
Verified locally: 225/225, and I checked the two things that could have gone wrong:

1. **Does the new signature get shadowed by `transient-network`?** No — I was worried because `isTransientPnpmFailure` already matches `ETIMEDOUT`/`network timeout` and sits earlier in the classifier, but you scoped the abort signature (`operation was aborted due to timeout` / error code 23 / `TimeoutError`) tightly enough that the two are disjoint: the abort shapes classify `fetch-timeout`, while `ETIMEDOUT`/socket-hangup/5xx still classify `transient-network`. Confirmed by running both sets through `classifyPnpmFailure`.
2. **Does pnpm actually accept the flag?** Yes — `pnpm add --config.fetchTimeout=600000` runs clean on 11.22.

The reasoning behind the separate recovery is the part I like most: a plain retry (the #83 path) would hit the same 60s wall, so the retry *must* carry a longer timeout to be worth anything. Reusing the `RELEASE_AGE_OVERRIDE` one-shot-override pattern keeps it consistent with the existing recoveries.

Thanks @Bruce-Yii — second solid one today.

Issue #113, opened 2026-08-16
[Bug] approve-and-retry still not shown for git prepare on 1.9.0: parsePrepareNotAllowed misses ndjson-escaped quotes


yzr278892:
### What happened / 发生了什么

On dshmarket **1.9.0** (which contains the #68/#69 fix), installing `dsh-queue-plus` still shows the old raw error instead of the approve-and-retry banner:

```
安装失败: dsh-queue-plus — " allowlist.
    at preparePackage (pnpm.mjs:137796:11)
    at async prepareGitHostedPkg (pnpm.mjs:138772:37)
...
```

The exported log shows the friendly classified message IS appended to stderr:

```
这个 git 插件需要在安装时执行构建脚本，被 pnpm 默认拦截。点击「允许构建脚本并重试」放行后重试即可 / ...
```

…but the client never shows the banner, and the user can only retry into the same raw error.

### Root cause / 根因

`parsePrepareNotAllowed()` uses:

```ts
const m = /git-hosted package "([^"]+)" needs to execute build scripts/.exec(`${stdout}\n${stderr}`)
```

With `--reporter=ndjson` (which the market always appends), pnpm emits the error as a JSON string whose embedded quotes are escaped. The raw captured stdout contains:

```
git-hosted package \"dsh-queue-plus@0.3.0\" needs to execute build scripts
```

The regex expects a literal `"`, so it returns `null`; `blockedBuilds()` then returns `undefined`, and the `/dsh-market/install` response has:

```json
{
  "ok": false,
  "exitCode": 1,
  "ignoredBuilds": null,
  "error": null
}
```

The client falls back to `stderr`/`stdout` and displays the raw stack. This is why the fix works for the human-line path but not the market's real pnpm 11 ndjson path.

### Repro / 复现

1. Install dshmarket 1.9.0.
2. Market → install `dsh-queue-plus` (`github:starslittle/dsh-queue-plus`, package has `prepare`).
3. Observe `/dsh-market/install` response: `exitCode=1`, `ignoredBuilds=null`, `error=null`, raw pnpm stack in `stdout`.
4. The exported log contains the friendly `git-prepare-not-allowed` classification, proving `classifyPnpmFailure()` matched but `blockedBuilds()` did not.

Minimal unit-level check against captured output:

```ts
parsePrepareNotAllowed(stdout, stderr) // => null
stdout.includes('git-hosted package \\"dsh-queue-plus@0.3.0\\" needs to execute build scripts') // => true
```

[...]

PR #119, opened 2026-08-16
feat(install): reclaim orphaned pnpm store staging dirs after failed runs

Bruce-Yii:
## Summary

pnpm stages every fetched tarball under the store's `tmp/` as `tmp/_tmp_<pid>_<hex>/`. A run that is **killed, cancelled, timed out, or hard-fails mid-fetch** never finishes staging, so the directory survives. For `github:` sources the staged payload is the **whole repository tarball**, so a single aborted install can leave hundreds of megabytes behind.

Real-world case that prompted this: an aborted install of `@openviking/dsh-memory-plugin` (`github:volcengine/OpenViking#path:/examples/dsh-memory-plugin`) left **~88MB** of orphaned staging data in the store tmp; the store accumulated 5 orphans (~94MB) over a few failed/cancelled attempts.

The staging name carries the owning pnpm pid, which makes reclamation **safe by construction**: only directories whose pid is no longer alive are removed, so a live download is never touched.

## Change

- **`src/store.ts`** (new): `cleanOrphanedStoreTmp(storePath)` removes dead-pid staging dirs; `cleanOrphanedStore(run, profile)` resolves the store root through the same runner installs use (`dsh plugin … store path` — so the web and Desktop pnpm paths agree) and reclaims orphans.
- **`src/install.ts`**: `withHoistRecovery` reclaims orphans whenever the final result is **not ok** (failed / timed out / cancelled) — successful runs clean up after themselves, so the happy path costs nothing extra.

## Tests

- `tests/store.spec.ts`: dead-pid staging dirs are removed; live-pid and non-staging entries survive; missing/relative store paths and runner failures are no-ops.
- `tests/install.spec.ts` / `tests/recovery.spec.ts` / `tests/flows.spec.ts`: failed runs now probe the store exactly once (updated call-sequence expectations).

Local: 227/228 tests pass. The single failure (`tests/backup.spec.ts` symlink creation) also fails on a clean checkout on this machine — Windows requires elevation to create symlinks, unrelated to this change.


fkysly (approved):
Approved after a careful look — deleting files needs it, and this is designed conservatively enough to earn it.

**Safety, verified end-to-end.** I built a fake store with four things in it and ran `cleanOrphanedStoreTmp` against it:

```
tmp/_tmp_99999999_dead   ← dead pid   → removed ✅
tmp/_tmp_<live-pid>_alive ← live pid  → untouched ✅
tmp/not-pnpm-dir          ← wrong shape → untouched ✅
files/keepme              ← store content → untouched ✅
```

The layering is what makes it sound: `storePath` comes from pnpm itself (`store path`) rather than being guessed, it must be absolute, only `_tmp_<pid>_` names are considered, the pid must be gone (with `EPERM` correctly treated as *alive*), and `rmSync` failures are swallowed so a locked dir just waits for a later run. Calling it only on terminal failures — after pnpm has exited — means nothing live can be mid-write.

**Merge resolution.** Your branch had unresolved conflict markers in `tests/install.spec.ts` (main moved twice under you today). I resolved it on your branch (`c747d99`): kept main's #113 test, grafted your store test, and updated one assertion in #117's "does not double-apply the fetchTimeout override" case — with #119 in place, a terminal failure now also probes `store path`, so the exact-call array grew by one entry. That is a real and intended interaction between your two PRs, not a regression. 230/230 after that.

Three good ones today, @Bruce-Yii — the 88MB real-world case made this easy to prioritize.

PR #150, opened 2026-08-17
fix: guard hotMount activation with a timeout so a wedged plugin cannot lock the market

Oo0520:
### For code changes / 代码改动

- [x] `npm test` and `npm run check` pass locally / 本地通过
- [x] Tests cover the change — for a bug fix, one that fails without it / 新增测试覆盖改动；bug 修复请确保测试在修复前是红的
- [x] No version bump in `package.json` — releases are a maintainer action / 不要改版本号，发版由维护者操作
- [x] `client/client.js` rebuilt with `npm run build:client` if you touched `src/client/` / 改动前端后重建产物 — server-only change, no client files touched

## Problem / 问题

`hotMount()` awaited `handle.await()` with no ceiling. When a freshly installed
plugin's activation never settles — typically a fiber pending on a service
nothing provides — the HTTP request hangs forever:

1. The route's `finally { installing = false }` never runs, so `installing`
   stays `true` and **every later install/update/uninstall is 409'd**
   ("another install is already running") until a host restart.
2. On the client it shows up as the infinite-spinner symptom (cf. #138).
3. Boot-time mounts (`mountClientOnlyDeps`) loop through `hotMount`, so a
   single stuck client-only package can hang the whole market startup.

This mirrors the route-level guard added for the Gist channel in #144 ("a
wedged gh CLI can never leave the client hanging") — the same class of
hazard, one layer down.

## Change / 改动

`src/hot.ts` (server-only):

- Race the activation against a **10s ceiling**, env-overridable via
  `DSH_MARKET_HOT_MOUNT_TIMEOUT_MS` (mirrors `INSTALL_TIMEOUT_MS` in
  `dsh-cli.ts`).
- On timeout: unwind the half-mounted subtree best-effort via
  `handle.dispose()` (disposal never blocks the reply), and fall back to the
  existing restart-required path with a `reason` that names the timeout.
- Handlers stay attached to the original promise, so a **late rejection
  after a timeout cannot surface as an unhandled rejection**.
- Non-timeout activation failures keep the previous behavior exactly.

## Test plan / 测试

`tests/hot.spec.ts`: a fiber whose `await()` never settles
(`new Promise(() => {})`), mounted through a fake context under fake timers.

- **Before this change**: the test times out after 20s — it hangs exactly
  like the route did.
- **After**: the mount reports `ok: false` with the timeout reason,
  `dispose()` ran, and the package is not listed as live.

Full local gates: `npm test` (355 passed, 1 skipped) and `npm run check`
(typecheck + build + restart-smoke) both green.

Related (not claiming to fix): #129 — the timeout reason feeds the same
"restart will fix it" messaging path this PR improves.

### DIFF
--- a/src/hot.ts
+++ b/src/hot.ts
@@ -38,6 +38,14 @@ interface HotContext {
 
 const HOT_DIR = '.dsh-market'
 
+/**
+ * Ceiling for one hot-mount activation, env-overridable like the install
+ * timeout in dsh-cli.ts. An activation that exceeds it is treated as wedged
+ * (typically a plugin pending on a service nothing provides) and falls back
+ * to restart activation.
+ */
+const HOT_MOUNT_TIMEOUT_MS = Number(process.env.DSH_MARKET_HOT_MOUNT_TIMEOUT_MS) || 10000
+
 let hotTreeClass: unknown | null | undefined
 
 /**
@@ -239,6 +247,26 @@ let hotSequence = 0
 
 const hotHandles = new Map<string, PluginHandle>()
 
+/** Activation did not settle within HOT_MOUNT_TIMEOUT_MS. */
+class ActivationTimeout extends Error {}
+
+/**
+ * Race an activation awaitable against the hot-mount ceiling. The handlers
+ * stay attached to the original promise, so a late rejection after a timeout
+ * can never surface as an unhandled rejection.
+ */
+function raceActivationTimeout<T>(awaitable: T | Promise<T>): Promise<T> {
+  return new Promise<T>((resolve, reject) => {
+    const timer = setTimeout(() => {
+      reject(new ActivationTimeout(`activation did not settle within ${HOT_MOUNT_TIMEOUT_MS / 1000}s — the plugin may be waiting on a service that never arrives`))
+    }, HOT_MOUNT_TIMEOUT_MS)
+    Promise.resolve(awaitable).then(
+      value => { clearTimeout(timer); resolve(value) },
+      error => { clearTimeout(timer); reject(error) },
+    )
+  })
+}
+
 /** Outcome of one hot-mount attempt; `reason` explains non-`ok` results. */
 export interface HotMountResult {
   ok: boolean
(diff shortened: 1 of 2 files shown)


fkysly (approved):
很有价值的一条，合了。358/358，我也把红→绿验证做实了。

## 为什么这个问题严重

正如你在描述里点明的：卡住的不只是一次安装，而是**整个市场**——路由的 `finally { installing = false }` 永远不执行，之后所有安装/更新/卸载一律 409「another install is already running」，直到重启宿主。这也解释了 [#138](https://github.com/dsh-market/dsh-market/issues/138) 那个「无限转圈」的现象。启动期的 `mountClientOnlyDeps` 会遍历 `hotMount`，一个卡住的 client-only 插件就能拖住整个启动流程，这点你也考虑到了。

## 我验证了什么

**红→绿是真的**。这里有个坑值得记一笔：我第一次用 `git stash push src/` 回退代码，测试照样通过——因为你的改动是已提交状态、不在工作区，stash 根本没动它。改用 `git show origin/main:src/hot.ts` 覆盖后重跑，测试**在 20 秒后超时失败**（正是「永远不返回」的病症），恢复后通过。

**实现细节我逐条看过：**

- `raceActivationTimeout` 把处理器挂在原始 promise 上，超时后迟到的 rejection 不会变成未处理拒绝 —— 这个细节容易漏，你没漏；
- 超时后 `dispose()` 半挂载子树，且用 `Promise.resolve(...).catch(() => {})` 包住，disposal 本身不阻塞回复；
- **`throw` 发生在 `hotHandles.set` 之前**，所以超时不会在 live 表里留下一条假记录——我特意确认了这个顺序；
- 超时可通过 `DSH_MARKET_HOT_MOUNT_TIMEOUT_MS` 覆盖，与 `dsh-cli.ts` 里安装超时的既有做法一致；
- 失败回落到重启激活，语义与其他热挂载失败路径统一。

顺带一提：我试图用真实 dsh web 造一个「卡死插件」来端到端复现，但 cordis 的 `inject` 是延迟满足而非阻塞挂载，造不出真卡死——你用假时钟 + 永不 settle 的 `await()` 直接打在 `hotMount` 边界上，是更准确的测法。

PR 模板的自查项你逐条勾了，也确实都属实（server-only 改动、没动版本号、测试红→绿）。感谢 @Oo0520。

PR #190, opened 2026-08-18
fix: spawn the real Node binary for children on Android (execPath is linker64)

DM0x32cc:
**What changed and why / 改动内容与原因**

Every plugin install/update/uninstall through the market failed on Android hosts with:

```
error: expected absolute path: "--expose-internals"
```

(exit 1, empty stdout — the child never ran).

**Root cause / 根因** — the market re-invokes the dsh CLI that launched the host by spawning `process.execPath` with `[...process.execArgv, <dsh bin>]`. On Android the kernel runs node through the dynamic linker, so `process.execPath` is `/apex/com.android.runtime/bin/linker64` while the real node binary is `process.argv0`. Spawning the linker with `--expose-internals <script> …` makes the linker treat `--expose-internals` as the program path — not an absolute path — and it exits with the error above. Verified by running the exact spawned argv:

```
$ /apex/com.android.runtime/bin/linker64 --expose-internals <dsh bin> plugin --profile web add <pkg>
error: expected absolute path: "--expose-internals"
```

**Fix / 修复** — new `nodeExecutable()` helper: prefers `process.argv0` when it is an existing absolute path, falls back to `process.execPath`. Used for:

- the dsh CLI re-invocation (`dshArgv`) — the failing path above;
- the self-restart helper (`restart.ts`) — same bug, the detached helper never started;
- `nodeBinDir` — was `dirname(execPath)`, a bogus `/apex/.../bin` on Android; now the real node dir so `npm`/`corepack` discovery keeps working.

No behavior change on macOS/Linux/Windows, where `argv0` is the same binary as `execPath`.

**Tests / 测试**

- `tests/dsh-cli.spec.ts`: new `nodeExecutable` cases — prefers an existing absolute `argv0` even when `execPath` is the linker (red before the fix), returns `argv0` verbatim, falls back to `execPath` for empty/relative/nonexistent `argv0`.
- `tests/provision.spec.ts`: the #167 fixture derives the node directory the same way the market does (`nodeExecutable()`), keeping it meaningful on Android while behaving identically on CI.
- Local run: 543/548 tests pass; the 5 `restart-helper.spec.ts` failures are environmental — that spec spawns `process.execPath` children itself and hits the same Android linker issue (identical failures on an unmodified checkout; they pass on CI). Typecheck (`tsconfig.json` + `tsconfig.tests.json`) is green.

**Checklist**

- [x] `npm test` / `npm run check` pass locally (see the environment note above; full `check` needs npm, which is unavailable on the Android test host)
- [x] Tests cover the change — the Android-preference test fails without the fix
- [x] No `src/client/` changes

### DIFF
--- a/src/dsh-cli.ts
+++ b/src/dsh-cli.ts
@@ -34,6 +34,23 @@ import { profileDir } from './profile.ts'
 const extraPathDirs: string[] = []
 
 /**
+ * The real Node executable for spawning children. On Android the kernel runs
+ * node through the dynamic linker, so `process.execPath` is
+ * `/apex/.../linker64` — spawning IT with `--expose-internals` makes the
+ * linker treat the flag as the program path and die with
+ * `error: expected absolute path: "--expose-internals"`. `process.argv0`
+ * carries the real node binary; prefer it whenever it is an existing
+ * absolute path, and fall back to execPath everywhere else.
+ * @param argv0 - `process.argv0`, injectable for tests.
+ * @param execPath - `process.execPath`, injectable for tests.
+ */
+export function nodeExecutable(argv0: string | undefined = process.argv0, execPath: string = process.execPath): string {
+  if (argv0 !== undefined && argv0 !== '' && isAbsolute(argv0) && existsSync(argv0))
+    return argv0
+  return execPath
+}
+
+/**
  * The directory holding the Node binary running this process. `npm`,
  * `npm.cmd` and `corepack` are installed alongside it by every official Node
  * distribution, so it is the one place the toolchain can be looked for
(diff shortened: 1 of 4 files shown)


fkysly (commented):
Pulled locally, ran the full suite (548 passing, matches main's count — nothing dropped) and `npm run check` — both green. Reviewed the diff: `nodeExecutable()` is a clean, minimal fix, and the guard (`isAbsolute` + `existsSync`) makes it safe to fall through to `execPath` everywhere it isn't the exact Android shape, so no behavior change risk on macOS/Linux/Windows — confirmed by the full suite passing unchanged.

Mutation-tested the new function myself: 2 of 3 mutants I tried are killed by your tests (removing the fix entirely, dropping the existsSync check). One survives — dropping `isAbsolute` alone isn't caught, since none of the "not absolute" test cases also happen to exist relative to the test runner's cwd. Not a correctness bug (the guard is still right), just a coverage gap — worth an extra case like `nodeExecutable('package.json', execPath)` (a real, existing, relative path) if you want to close it, but not blocking.

The CI run needed maintainer approval (standard gate for a first PR from a new contributor to this repo) — approved it, running now. Will merge once it's green.

PR #196, opened 2026-08-18
fix: polish market UI search, installed view, and sticky categories

chengexin:
## What

- Make search fields span the full row on Discover, Themes and Installed.
- Move the Installed list/groups toggle below the search box and align right.
- Keep category chip order fixed after selection.
- Remove body top padding so the sticky category bar covers scrolling content.
- Normalize sticky category bar padding to 4px on all sides.

## Why

These are UI issues found in normal usage:

- Search box width was inconsistent across pages.
- Installed page list/groups toggle was above the search box and left-aligned.
- Clicking a category tag moved the tag position.
- When the category bar stuck to the top, content could be seen through the gap above it.


fkysly (changes requested):
Thanks for this — pulled it locally, ran the full suite (545 passing) and `npm run check` clean, and real-host verified each piece.

**The search-full-width, installed-view-toggle repositioning, and sticky-bar padding tweak are all good** — verified on a real host, no issues, happy to take these as-is.

**The category-chip fix has a real regression, verified on a real host with a screenshot.** `orderedCategories` now returns catalog order unconditionally — never rescuing the active chip to the front — which removes the entire reason that mechanism existed: when the row is collapsed, only the first `visibleCats` chips render at all (`ordered.slice(0, visibleCats - 1)`), so a category selected while the row was expanded (or picked from further down before it was ever collapsed) can end up outside that slice. I reproduced it: expanded the row, picked "Just for Fun" (last chip), collapsed — the grid correctly filters to "Just for Fun" plugins, but **no chip anywhere in the collapsed row shows as selected**. Screenshot attached to this review shows it — a user has no way to tell which category is active.

This was independently reported to us by a screenshot the same day (the annoying-reorder half of what you found), and I shipped a fix in #197 that keeps the rescue but makes it conditional — only reorder when the chip's natural position would actually fall outside the visible budget, so an already-visible chip never moves (fixing exactly what you're both reporting) while a genuinely-hidden one still gets rescued (avoiding the regression above). Mutation-tested, plus a real-host e2e for both directions.

Once #197 lands, could you rebase and drop the `orderedCategories`/`market-data.spec.ts` hunk from this PR? The other three changes (search width, toggle position, padding) don't overlap with anything and I'd merge those on their own — happy to take a follow-up PR with just those if that's easier than rebasing this one.

fkysly:
先说一声：1.15.0 里这块区域改动比较大，可能会跟这个 PR 冲突，麻烦看下要不要一起更新一下——

- 搜索框滑动时被裁切的问题，换了个思路修的：不是调 padding，是把搜索框和分类栏包进同一个 sticky 容器，两者一起吸顶（详见 `.stickyHead`，src/client/Market.module.css）。这个和你 PR 里"去掉 body 顶部 padding 让分类栏盖住内容"的方案思路不一样，需要确认一下叠加后效果对不对，或者看看这个 PR 里还有哪些部分是独立于这块的（比如已安装页 list/groups 开关位置、分类 chip 顺序不跑位那两点，这些看起来跟吸顶没关系，可以分开处理）。
- `.cats` 现在也不再单独 `position: sticky` 了，改成外层 `.stickyHead` 统一处理，如果 PR 里对 `.cats` padding 的调整还需要，可能要挪到新结构里。

不用着急，等你方便的时候看一下，需要的话可以拆成几个小 PR 分别处理。升级到 1.15.0 之后可以实际感受下现在的效果。

PR #217, opened 2026-08-19
fix: restore in-flight update progress when the config page reopens

ericw0315:
## What & why

Closing the Settings → Plugin Market page while a plugin update is running **loses the update progress**. Reopening the page shows no running row — even though the update is still executing server-side — because the running row is driven by `updatingName`, a React state in `MarketSection.tsx` that dies with the component on unmount.

The install flow already survives this via a `dshm-pending` sessionStorage marker + the `/dsh-market/status` poll recovery; the update flow had no equivalent.

## Fix

Mirror the install flow for updates:

1. `doUpdate` writes a `dshm-updating` sessionStorage marker before firing the request.
2. The marker is cleared once the HTTP response arrives (any 2xx/4xx/5xx).
3. A **lost** response (`.catch`) keeps the marker and the running row instead of declaring a false failure — same reasoning as the install flow's #100 recovery.
4. On mount, `updatingName` is restored from the marker.
5. The status poll's idle branch hands the restored row back (clears marker + `updatingName`, refreshes the installed listing) once the host reports the operation settled — pnpm exited AND the mutation lock released — guarded by the same two-idle-poll grace as the install #32 recovery.

## Changes

- `src/client/MarketSection.tsx`
- `tests/client/market-section.client.spec.tsx`

## Test plan

- Added regression test: `lost update progress (config page reopened)` — restores the running row + live progress from the marker, then converges once the host settles.
- `tests/client` + `market-data.spec.ts`: **124 tests pass**.
- `tsc -p tsconfig.client.json --noEmit`: clean.

### DIFF
--- a/src/client/MarketSection.tsx
+++ b/src/client/MarketSection.tsx
@@ -351,6 +351,8 @@ export function MarketSection(props: MarketSectionProps) {
   const [busyUrl, setBusyUrl] = useState<string | null>(null)
   /** Consecutive idle polls with a pending install that never landed (#32). */
   const idleStrikes = useRef(0)
+  /** Same idle-strike bookkeeping for an update whose response was lost. */
+  const updateIdleStrikes = useRef(0)
   const [doneUrls, setDoneUrls] = useState<string[]>([])
   const [installError, setInstallError] = useState<string | null>(null)
   interface CompatibilityNotice {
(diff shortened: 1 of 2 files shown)


ericw0315:
使用fix分支重新提交，此PR close

PR #237, opened 2026-08-20
fix: ignore negated terminal descriptions

WSL043:
Fixes the compatibility warning shown for web plugins whose descriptions say a CLI is not required.

The terminal heuristic now ignores bounded negated mentions such as "no CLI required" and "无需 CLI", while package names and positive terminal/TUI descriptions remain strong signals.

Validation:
- Added regression coverage that fails before the fix and passes after it.
- Focused market-data suite: 35/35 passed.
- Rebuilt client/client.js as required.
- The broader local run reached 629 passing tests and 2 skipped; two unrelated suites could not load because the current package metadata does not declare @deepseek-ai/schemastery.


fkysly:
已合并并随 **1.16.3** 发布，感谢贡献 🙏

```
```

（合并前我 rebase 到了最新 main——冲突来自我这边刚改过的同一批文件，你的改动本身没有动过。）

PR #250, opened 2026-08-21
fix: install prebuilt registry tarballs

ruisenbai:
## Summary

- honor the curated `tarball` field after a repo-verified npm package and before GitHub source fallback
- accept only HTTPS GitHub Release hosts with `.tgz` or `.tar.gz` archives
- document the prebuilt install path and cover its priority and rejection cases

## Why

`awesome-dsh-plugin` already exposes author-supplied prebuilt Release tarballs. dsh-market ignored that field, so GitHub-only packages such as `dsh-inline-comments` fell back to a source checkout and required pnpm build approval even when a ready-to-install archive existed.

## Checks

- `npm ci` (includes production build)
- `npm run typecheck`
- `npm test -- tests/sources.spec.ts`

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ its own dsh: it may be older than the one `npm` would give you (#139).
 
 ## Speed
 
-Installs prefer npm tarballs over full-repo GitHub downloads whenever a plugin publishes to npm (registry-verified against the repo to prevent name squatting). Registry installs are typically seconds; GitHub-only plugins depend on your connection to GitHub.
+Installs prefer repo-verified npm packages, then author-supplied prebuilt GitHub Release tarballs, before falling back to full-repo GitHub source downloads. Prebuilt installs are typically seconds and do not need local build scripts; source-only plugins depend on your connection to GitHub.
 
 ## Security
 
--- a/README.zh.md
+++ b/README.zh.md
@@ -51,7 +51,7 @@ dsh plugin --profile web add dshmarket
 
 ## 速度
 
-只要插件发布了 npm 包（registry 会校验其 repository 指回同一仓库,防冒名）,安装即走 npm tarball 而非整仓 GitHub 下载——通常秒级;仅 GitHub 分发的插件取决于你到 GitHub 的网络。
+安装依次优先使用经仓库验证的 npm 包、作者提供的 GitHub Release 预构建 tarball，最后才回退到整仓 GitHub 源码下载。预构建安装通常只需数秒且无需执行本地构建脚本；仅提供源码的插件仍取决于你到 GitHub 的网络。
 
 ## 安全
 
--- a/src/client/market-data.ts
+++ b/src/client/market-data.ts
@@ -15,6 +15,7 @@ export interface RegistryPlugin {
   owner: string
   url: string
   npm?: string
+  tarball?: string | null
   category: string
   description?: LocalizedText
   stars?: number
--- a/src/registry.ts
+++ b/src/registry.ts
@@ -13,6 +13,7 @@ export interface RegistryPlugin {
   category: string
   description: Record<string, string>
   npm?: string | null
+  tarball?: string | null
   stars?: number | null
   /**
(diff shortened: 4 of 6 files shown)


fkysly (changes requested):
Good change and the right priority order — thanks. The protocol, host and extension checks are all solid; I probed `releaseTarballTarget` directly and `http:`, non-`/releases/` paths, `.zip`, and off-allowlist hosts are all correctly rejected.

One gap I'd like closed before merging.

## The tarball is not bound to the entry's own repo

`installTargetFor` verifies the `npm` field against the repo — the docstring calls it "name-squatting protection" — but the new `tarball` branch has no equivalent. A catalog entry can name one repo and install an archive from a different one:

```js
// entry.url    = https://github.com/good/plugin
// entry.tarball= https://github.com/evil/repo/releases/download/v1/p.tgz
releaseTarballTarget(entry.tarball)  // → accepted
```

I ran that: it passes. Same for `https://objects.githubusercontent.com/whatever/x.tgz`, which carries no owner/repo in its path at all.

This is exactly the class the npm branch already guards against, so the tarball branch should hold the same line. `parseSourceUrl(entry.url)` already hands you `source.repo`, so for a `github.com` URL it's a comparison of the first two path segments.

## On the two CDN hosts

`objects.githubusercontent.com` and `release-assets.githubusercontent.com` are what GitHub *redirects* release downloads to; a curated entry would normally carry the `github.com/<owner>/<repo>/releases/download/...` form, and pnpm follows the redirect itself. Since the owner/repo can't be recovered from those paths, they can't be bound to the entry either — so I'd suggest dropping them and accepting only the `github.com` release form. If you have a real catalog entry that needs them, say so and we'll treat that as unverifiable-by-design with a comment saying why.

Happy to merge once the tarball is required to come from the entry's own repo. A test for the cross-repo rejection alongside the existing ones would be good to have.

fkysly:
把我 review 里提的那个缺口补上了，直接推到了这个分支（你开了 maintainer 可编辑）——你的提交原样保留，我的加在后面。

**补了什么**

`releaseTarballTarget` 现在多收一个参数：条目自己的 `owner/repo`，并要求 tarball 路径的前两段与之相等（大小写不敏感，跟 GitHub 一致）。这样 npm 分支上那句"repo-verified，防冒名"在 tarball 分支上也成立了。

顺带把 `objects.githubusercontent.com` 和 `release-assets.githubusercontent.com` 两个 CDN 主机从白名单里去掉了。不是因为它们不安全，是因为**它们的路径里根本没有 owner/repo**——没有可绑定的东西，也就无法判断这个包是谁的 release。留着它们等于把同一个洞换个名字继续开着。

**这会不会误伤真实条目**

不会，这是我实测过才敢收紧的。目录里 2062 条，带 `tarball` 的 70 条：

- 全部使用 `github.com`（没有一条走 CDN 主机）
- 全部的 owner/repo 与条目自身一致（0 条不匹配）
- 收紧后：51 条仍走 tarball，19 条本来就因为有 npm 包而优先走 npm，**因收紧而退回源码下载的：0 条**

**测试**

你原来那条断言 CDN 主机被接受的用例，我改成了断言它被拒绝，并补了两条：一条覆盖"是真实 Release 归属却不对"（含 `evil/repo`、`o/other`、`evil/r` 和两个 CDN），一条覆盖大小写差异和 monorepo 条目（`/tree/main/packages/x` 仍按 repo 绑定，不按子路径）。

618 个测试全过，typecheck 干净。功能方向本来就是对的——预构建包免掉本地构建授权，这是实打实的体验改进，谢谢你做这个。等 CI 绿了我就合。

fkysly (approved):
缺口已在本分支补齐（见上一条评论），实测 70 条真实 tarball 条目零回退。

fkysly:
已合并，随 **1.22.0** 发布。谢谢 @ruisenbai——预构建 tarball 省掉本地构建授权这一步，对 GitHub-only 插件是实打实的体验提升。

PR #263, opened 2026-08-22
fix: reach the catalog through a proxy that lives only in npm config

tianguoyan:
## The catalog failed on a machine whose proxy lives only in npm config

**Symptom**: the market shows `fetch failed (21s, 2 attempts)` although every npm-based tool on the machine works. The proxy was configured with `npm config set proxy` (common on Windows): it lives in `npm_config_proxy` / `npm_config_https_proxy` (and `.npmrc`), but not in `HTTP_PROXY` / `HTTPS_PROXY`. The catalog fetch therefore took the direct route and timed out at the same GitHub Pages IPs that prompted #187.

**Cause**: `configuredProxy()` resolved only the standard `http(s)_proxy` variables, so on such a machine it returned `null` and `marketFetch` fell back to Node's global `fetch` — exactly the direct path `EnvHttpProxyAgent` was introduced to avoid.

**Fix** (`src/net.ts`):
- `proxyFromEnv()` resolves the standard variables first, then falls back to `npm_config_https_proxy` / `npm_config_proxy` (same precedence as undici: lowercase wins, https falls back to http, blank is unset; a standard variable always beats npm's config).
- `marketFetch` builds `EnvHttpProxyAgent` with the resolved proxies passed explicitly (`httpProxy` / `httpsProxy`). `EnvHttpProxyAgent` reads only `http(s)_proxy` out of the environment, so without the explicit handoff the npm fallback would be a false claim in the failure message while the agent still went direct.

**Tests** (`tests/registry.spec.ts`), red before green:
- `configuredProxy` npm-fallback cases (https, http fallback, standard precedence, empty-is-unset). The per-test env cleanup now covers `npm_config_*` too, so the suite stays deterministic when the developer's own `.npmrc` carries a proxy.
- `marketFetch` (undici mocked) pins that npm-sourced proxies reach the agent explicitly, and that the global-fetch path is kept when no proxy is configured anywhere.

With the source fix stashed the new tests fail — the npm fallback returns `null` and the handoff test reproduces the direct `ENOTFOUND`.

Verified end-to-end on the reporting machine: `GET /dsh-market/registry` went from 502 `fetch failed (21s, 2 attempts)` to 200 (full catalog) with the equivalent local change.

Known non-goal: `npm_config_noproxy` is deliberately not consulted — `no_proxy` behavior stays exactly as `EnvHttpProxyAgent` defines it.

Local checks: `tsc -p tsconfig.tests.json --noEmit` / `tsc -p tsconfig.json --noEmit` clean; `tests/registry.spec.ts` 30/30. The full unit suite shows 15 failures in `dsh-cli`/`flows`/`updates` specs that are identical on a clean checkout (live npm/GitHub version assertions, unrelated to this change).

### DIFF
--- a/src/net.ts
+++ b/src/net.ts
@@ -25,11 +25,16 @@ import { EnvHttpProxyAgent, fetch as undiciFetch } from 'undici'
 /**
  * The proxy this process would use for the catalog, if any.
  *
- * This mirrors `EnvHttpProxyAgent`'s own resolution deliberately, rather
- * than picking the order that reads best, because the same answer does two
- * jobs: it decides whether to route through undici at all, and it is what
- * the failure message CLAIMS was tried. A helper that named a proxy undici
- * would not have used would put a false statement in every bug report.
+ * The standard variables mirror `EnvHttpProxyAgent`'s own resolution
+ * deliberately, rather than picking the order that reads best, because the
+ * same answer does two jobs: it decides whether to route through undici at
+ * all, and it is what the failure message CLAIMS was tried. A helper that
+ * named a proxy undici would not have used would put a false statement in
+ * every bug report. `npm_config_*` is an additional source on top of that:
+ * npm holds its proxy in its own config namespace (a machine set up with
+ * `npm config set proxy` has the proxy in `npm_config_proxy` and nowhere
+ * else), and undici does not read it — so it is resolved here and handed to
+ * the agent explicitly in `marketFetch`.
  *
  * Three details are undici's, not ours (env-http-proxy-agent.js):
  *   - lowercase wins over uppercase (`https_proxy ?? HTTPS_PROXY`)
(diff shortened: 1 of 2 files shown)


fkysly:
Verified the gap and it's real: `configuredProxy()` (`src/net.ts:43`) reads only `https_proxy`/`HTTPS_PROXY`/`http_proxy`/`HTTP_PROXY`. It never reads `npm_config_*` and never reads `.npmrc`, so a machine configured with `npm config set proxy` gets `null` back and `marketFetch` falls through to the global `fetch` — exactly the direct route `EnvHttpProxyAgent` exists to avoid. Good catch, and it's the symmetric counterpart to the `HTTPS_PROXY` → `npm_config_*` translation that went the other way for the pnpm subprocess in 1.16.2.

I checked the one thing I expected to be a problem and it isn't: passing `{ httpProxy, httpsProxy }` explicitly does **not** disable NO_PROXY handling, because undici resolves it as `this.#opts.noProxy ?? this.#noProxyEnv` and you don't pass `noProxy`. So "a host that excludes its own registry mirror keeps being excluded" still holds. Approved the workflow run.

One question before I merge — the PR body says the fix covers npm config "(and `.npmrc`)", but the code reads `process.env.npm_config_*` only. Those variables are exported by **npm when it spawns the process**; a directly-launched `dsh web` sees only the `.npmrc` file and no `npm_config_*` in its environment. You report an end-to-end fix, so your launch path clearly does export them — could you say how dsh is started on that machine? I'd rather the description match the code, either by narrowing the wording or by adding an actual `.npmrc` read.

PR #269, opened 2026-08-22
fix(diagnostics): guard the AI-fix prompt against fixing the running harness

skiuniverse:
## Problem

The advanced-diagnostics **AI fix** copies a diagnostics-driven prompt to the clipboard; the user pastes it into an agent and decides whether to send. The prompt never told the agent that the target profile is the **live profile the harness is running from**.

If the receiving agent is itself that harness, any in-place fix — reordering bundles, editing `cordis.patch.yml`, disabling a bundle, reinstalling deps, or upgrading / restarting the harness or a core package — mutates the very composition the agent is running from. The harness can re-compose on a patch change, falling back to a restart that kills the running process, so a fix executed inside the harness could crash it mid-fix.

## Change — self-identify first, then hand scripts to run externally

The prompt now makes the agent **detect whether it is itself the target harness** (`aiFixDetect`; when unsure, assume the safe answer), and if so (`aiFixIfSelf`) it:

1. does read-only analysis only, changing nothing;
2. produces an idempotent `apply.sh` / `.bat` (this fix's changes) plus `rollback.sh` / `.bat` (fully reverts it); the snapshot is taken **once only** on the first run, so a re-run can't clobber the pristine original, and rollback restores from that snapshot, so a half-run apply and repeated rollbacks both stay safe;
3. has the user run `apply` in an **external terminal** and paste `stdout` / `stderr` back, then judges / tunes / triggers `rollback`.

In-place mutation of the live composition / harness / core packages stays forbidden. `aiFixScope` is now conditioned on **not** being the target harness.

`zh` and `en` mirror 1:1; the rebuilt `client/client.js` reflects it.

## Also fixes a pre-existing build blocker

`src/settings.ts` imports `@deepseek-ai/schemastery` but never declared it, so a standalone `pnpm install` left it out of root `node_modules` and the host build / tests failed with `TS2307` (CI only worked because `npm install` hoists transitive deps). Added it as a devDependency and regenerated `pnpm-lock.yaml` (which was also stale relative to `package.json` — 7 declared deps were missing).

## Verification

- `npm run check` ✓ (typecheck host + client + tests, build, restart-smoke)
- `npm test` ✓ **706 passed**
- `preflight` / `validate-registry` / `smoke-spawn` ✓
- New regression assertions in `tests/client/diagnostics-panels.spec.tsx` verify the guard strings are present in the prompt.


fkysly:
已合并，随 **1.23.0** 发布（走的是 #307——你的分支带一个 merge commit，没法直接 cherry-pick，所以取净 diff 重放到当前 `main`，改动本身没变）。

这个问题找得很好。提示词把一个 profile 交给 agent 去修，却从没告诉它**这个 profile 可能就是它自己正跑在里面的那个**——真按它做（改 bundle 顺序、写 cordis.patch.yml、重装依赖、重启 dsh），修到一半把自己干掉了。

尤其认可默认值那一条：判断不了就按「是你」处理。**两个分支里只有这个方向猜错了也不会造成损坏**，这是对的取舍。

**我补了一处。** schemastery 你只加进了 `devDependencies`。这修好了本仓库的构建，但修不了用户那边：`src/` 在 `files` 里、是发布内容，`settings.ts` 又被 `index.ts` 静态引入，所以用户在严格 node_modules 布局下装了 dshmarket 之后，运行时照样解析不到。

已改成同时声明为 **optional peer**——跟同一个文件里、同样用法、同样来自宿主家族的 `@deepseek-ai/dsh-settings` 保持一致（顺带一提，`dsh-settings` 自己对 schemastery 的声明也正是 `peers: ^3.18.1`）。

823 测试全过。谢谢。

PR #315, opened 2026-08-24
fix: guard uninstall against user patch references

Jstn-1g:
## Summary

- inspect the profile's user cordis.patch.yml with the same YAML dialect DSH uses to compose it
- refuse uninstall when an insert entry still names the exact package or one of its exported subpaths, including nested loader groups
- fail safely when the existing patch is unreadable, malformed, structurally indeterminate, or cyclic
- leave the user-authored patch byte-for-byte untouched and avoid neighbouring package-name false positives

## Why

This addresses the uninstall-safety half of #165. The uninstall route currently reaches package removal without checking whether the profile's user patch still loads that package. Removing it can therefore leave an unresolved loader entry and break the next DSH boot.

The broader post-install activation classification remains separate, as agreed in the issue discussion.

## Verification

- focused patch, flow, and parser suites: 168/168 passed
- full suite: 917 passed, 1 skipped across 47 files
- type checking: all three TypeScript projects passed
- production build and restart smoke: passed
- git diff --check

### DIFF
--- a/src/check.ts
+++ b/src/check.ts
@@ -204,17 +204,20 @@ function isRecord(value: unknown): value is Record<string, unknown> {
   return value !== null && typeof value === 'object' && !Array.isArray(value)
 }
 
-/** Parse one entry-list patch file with the dsh dialect; null when unreadable. */
-export function parsePatchFile(path: string): unknown[] | null {
-  let text: string
-  try {
-    text = readFileSync(path, 'utf8')
-  } catch {
-    return null
-  }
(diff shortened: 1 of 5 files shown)


fkysly:
已合并，随 **1.27.0** 发布。谢谢 @Jstn-1g——这三个 PR 都是先把边界讲清楚再动手的那种，尤其是明确写出"只查包根存在性、不解析 exports"这类自我设限，比多修一点更有价值。

PR #316, opened 2026-08-24
fix: respect host-provided profile bundles

SparkElf:
## Summary

- treat active profile Bundles as installed only when matching Discover catalog cards
- keep update, uninstall, backup, and group operations on profile-owned dependencies
- report whether the profile owns the market dependency and hide self-update, channel, and removal controls when a host provides it

## Problem

A dsh distribution can provide Bundles from its application dependency closure while the profile manifest keeps `dependencies` empty. Those Bundles are active through `dsh.profile.bundles`, but Discover currently offers them for installation. Adding them directly to the dependency-owned installed map avoids that button but incorrectly enables update and uninstall, creating a profile-local shadow of the host-provided version.

The installed endpoint already returns both the dependency map and active Bundle names. This change keeps their ownership distinct and combines them only for catalog matching.

## Verification

- `pnpm exec vitest run tests/market-data.spec.ts tests/client/settings-card.client.spec.tsx tests/flows.spec.ts` (177 tests passed)
- `pnpm run typecheck`

Coverage includes dependency-spec precedence, host-provided catalog presence, status ownership for both profile-managed and host-provided installations, and hidden self-management controls.

### DIFF
--- a/src/client/MarketSection.tsx
+++ b/src/client/MarketSection.tsx
@@ -40,7 +40,7 @@ import { clearSettled, drop, enqueue, patch as patchRecord, recordForUrl } from 
 import type { OperationRecord } from './operations.ts'
 import { Diagnostics } from './Diagnostics.tsx'
 import {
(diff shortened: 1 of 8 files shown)


fkysly:
已合并，随 **1.28.0** 发布。

PR #318, opened 2026-08-24
fix: report unresolved user patch loader packages

Jstn-1g:
## Summary

- report effective user/home patch inserts whose npm package root is absent from the profile-visible `node_modules` ancestry
- preserve Loader activation semantics across disabled rows, conditional `!!js` values, nested groups, layer overrides, and orphan inserts
- keep the public `CheckReport`/client shape and restore behavior unchanged

## Why

The existing analyzer catches unresolved bundle packages, but a validly composed user/home patch can still reference an npm package root that is absent and fail only later when the Harness Loader boots. This adds that missing diagnostic to `/dsh-market/check`.

## Resolution boundary

This intentionally checks npm package-root presence, not exact `exports` or subpath resolution. It normalizes scoped/unscoped subpaths to their root, searches only the profile-visible `node_modules` ancestry, honors nearest-package shadowing and profile self-reference, and excludes path, scheme, builtin, and `#imports` specifiers.

## Regression coverage

- missing user and home packages, plus per-layer deduplication
- profile parent fallback, install-only exclusion, nearest-directory shadowing, file-shadow fallback, and profile self-reference
- nested group ownership and inserts targeting bundle-provided groups
- orphaned, disabled, conditional, and custom-group entries
- invalid/missing names and path/scheme/builtin/`#imports` exclusions
- Node-resolvable legacy and Unicode package roots
- malformed patch behavior

## Verification

Exact head: `a69e2b9b6cc7f84e69f338ff49746029109d9397`

- Linux, Node 24/npm 11: `npm test` — 47 files, 929 tests passed
- Linux, Node 24/npm 11: `npm run check` — type checks, production build, normalization, and restart smoke passed
- Linux, Node 24: `npm run test:compat` — real pnpm 9/10/11 matrix, 9 tests passed
- Windows: full suite and `npm run check` passed
- Pre-fix base `53cb827b12fad5021d9ccdecd1a2797f537e712d` with the exact regression file: 11 failed, 31 passed

## Scope

This implements only the standalone `/dsh-market/check` analyzer half described in #205. Restore-route surfacing remains separate, so this PR does not close the issue.

Refs #205

- [x] `npm test` and `npm run check` pass locally
- [x] Tests cover the change and fail without it
- [x] No `src/client/` changes; no client artifact update required

### DIFF
--- a/src/check.ts
+++ b/src/check.ts
@@ -18,16 +18,18 @@
  *     version (tool calls die, minimal preset fails to mount)?
  *  4. Are there multiple versions of one core package in the lockfile, and
  *     do plugin peerDependencies ranges match the resolved core version?
+ *  5. Do effective user/home patch entries reference npm package roots that
+ *     are installed in the profile-visible node_modules ancestry?
  *
  * The composition step mirrors @deepseek-ai/dsh-app-boot's applyEntryPatches
  * (same js-yaml dialect incl. `!!js` scalars), so the rows reported here are
  * what actually mounts at boot.
  */
 
-import { existsSync, readdirSync, readFileSync } from 'node:fs'
-import { createRequire } from 'node:module'
+import { existsSync, readdirSync, readFileSync, statSync } from 'node:fs'
+import { createRequire, isBuiltin } from 'node:module'
 import { homedir } from 'node:os'
-import { dirname, join, resolve } from 'node:path'
+import { dirname, isAbsolute, join, resolve } from 'node:path'
 import { JSON_SCHEMA, Type, load } from 'js-yaml'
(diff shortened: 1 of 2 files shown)


fkysly:
已合并，随 **1.27.0** 发布。谢谢 @Jstn-1g——这三个 PR 都是先把边界讲清楚再动手的那种，尤其是明确写出"只查包根存在性、不解析 exports"这类自我设限，比多修一点更有价值。

Jstn-1g:
Thank you for carrying all three into 1.27.0 and for the thoughtful refinement. The boundary-first approach is exactly what I want to keep bringing: fail only on evidence, preserve unknowns instead of manufacturing certainty, and keep recovery paths deliberate. Future contributions will stay similarly scoped and test-backed.

PR #323, opened 2026-08-25
feat: support multiple searchable categories

### LINKED ISSUE #321: feat: support multiple searchable categories per plugin
## Problem

Registry entries currently expose a single `category: string`. A plugin can reasonably belong to more than one existing catalog label, for example UI Enhancements plus Usage & Billing, or Skills plus Models & Providers. The current model forces registry maintainers to choose one label, and free-text search cannot discover a plugin through its category labels.

## MVP

Keep the existing category/chip/filter system and make it multi-value instead of introducing a second taxonomy:

- accept both legacy `category: string` and multi-value `category: string[]` in registry data
- normalize both forms to one internal category list
- render every assigned category with the existing card chip style
- include a plugin in a category filter when any assigned category matches
- include raw category ids and localized category labels in free-text search
- preserve existing name, owner, and description search behavior

Existing single-category registry entries must continue to work unchanged. Category ids should be de-duplicated while preserving declaration order.

## Acceptance criteria

- a fixture with two categories renders both existing chips
- selecting either category includes the plugin
- searching either category id or its localized label finds the plugin
- legacy string-category fixtures retain their current UI and filtering behavior
- no new installer, extension kind, capability schema, or separate tag-management surface is introduced

This also lets a skill pack remain an ordinary dsh plugin and use a Skills category together with other relevant categories.

SparkElf:
## Summary

- accept legacy `category: string` and multi-value `category: string[]` catalog entries, normalizing them to a de-duplicated category list in declaration order
- render every category with the existing chip UI and include a plugin when any assigned category matches the active filter
- search raw category ids and every localized category label alongside the existing name, owner, and description fields
- preserve multi-category theme lifecycle behavior in host routes and keep the static catalog and offline registry gate compatible

## Compatibility

Existing single-category catalogs continue to load unchanged. Malformed category members are omitted at the external registry boundary, and a plugin with no usable category is rejected instead of reaching the UI.

No installer, plugin kind, capability schema, or second tag taxonomy is introduced.

## Verification

- focused Vitest suites: 283 tests passed across registry loading, market-data filtering/search, theme classification, the real MarketSection component, and host flows
- `pnpm run check` (typecheck, host/client build, restart smoke)
- `pnpm run validate:registry`
- `pnpm run build:site`
- `git diff --check`

Closes #321


fkysly:
已合并，随 **1.28.0** 发布。

PR #344, opened 2026-08-25
fix: roll back profile bundles after failed plugin adds

### LINKED ISSUE #339: 通过设置页插件市场上更新版本经常导致dsh都启动不了
### 升级复测 / Retested on the latest version

- [x] 我已升级到最新版并重启，问题仍然存在 / I upgraded to the latest version, restarted, and it still happens

### dshmarket 版本 / version

1.29.0

### 运行环境 / Where are you running it

普通 dsh web（浏览器打开）/ plain dsh web in a browser

### 桌面端版本与来源 / Desktop build and version

0.1.1-rc.2

### 发生了什么 / What happened

本身装了不少插件，然后，通过设置页插件市场上更新版本经常导致dsh都启动不了。
已出现过2~3次，不知道日志在哪里，需要提供哪些进一步辅助定位的信息？

### 复现步骤 / Steps to reproduce

1. 打开设置 - 插件市场
2. 点击插件市场上的更新按钮，预期是更新插件市场到最新版

### 导出日志 / Exported log

[dsh-market-log.txt](https://github.com/user-attachments/files/31422819/dsh-market-log.txt)

### 操作系统 / OS

windows 11

Jstn-1g:
## What this fixes

A failed `dsh plugin add` can update both `dependencies` and `dsh.profile.bundles` before a later fetch or build-script failure. The market already restored the dependency map, but the newly added bundle row survived. On the next restart, the host then tried to activate a package that was neither installed nor declared as a dependency, bricking the profile described in #339.

## Changes

- snapshot the exact dependency map and the presence/value/order of `dsh.profile.bundles` before install and update operations
- restore both fields atomically after non-cancelled failures while preserving every unrelated manifest field
- carry the same snapshot through clean-exit update and Git rollback paths
- reproduce the host's partial bundle write in the full install-flow test and verify the pre-existing bundle list is restored exactly
- cover existing bundle lists, absent bundle fields, unrelated fields, and idempotent restoration at the profile layer

## Safety boundary

This does not perform a broad orphan sweep or remove user-authored bundle entries. Rollback is limited to restoring the exact values captured immediately before the operation. Cancelled operations retain their existing partial-state behavior.

## Verification

- `pnpm exec vitest run tests/profile.spec.ts tests/flows.spec.ts` — 152 passed
- `pnpm test` — 955 passed, 1 skipped
- `pnpm run typecheck` — passed
- build — passed

Fixes #339

### DIFF
--- a/src/profile.ts
+++ b/src/profile.ts
@@ -7,6 +7,7 @@
 import { existsSync, readdirSync, readFileSync, realpathSync, renameSync, statSync, writeFileSync } from 'node:fs'
 import { homedir } from 'node:os'
 import { dirname, isAbsolute, join, relative, resolve } from 'node:path'
+import { isDeepStrictEqual } from 'node:util'
 import { githubRemoteIdentities, githubRepoIdentities } from './sources.ts'
 
 /**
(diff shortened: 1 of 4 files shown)


fkysly:
已合并，随 **1.30.0** 发布。谢谢！

PR #368, opened 2026-08-27
fix: explain missing tarball integrity

### LINKED ISSUE #367: classifyPnpmFailure misses ERR_PNPM_MISSING_TARBALL_INTEGRITY (raw stack instead of actionable error)
## Summary

When a tarball-URL dependency in a profile's `pnpm-lock.yaml` has a `resolution` with `tarball:` but no `integrity:` field, pnpm 11 fails **every** subsequent install/uninstall with `ERR_PNPM_MISSING_TARBALL_INTEGRITY` (it replays the whole dependency tree). The market has no classifier for this code, so the user sees only dsh's wrapper line plus a raw pnpm stack tail — the exact "UI shows a stack tail and nothing else" symptom already called out in #244/#192/#138 — instead of an actionable, bilingual message.

## Repro

Observed on dsh-market 1.29.2, pnpm 11.7.0, Node 24.15.0, Windows, region `china`.

1. A profile has a plugin installed from a GitHub tarball URL. In the China region the market's acceleration rewrites `github:owner/repo` to a commit-pinned `https://gh-proxy.com/https://codeload.github.com/…/tar.gz/<sha>` URL, e.g.:

   ```json
   "dsh-think-translate": "https://gh-proxy.com/https://codeload.github.com/UncleK/dsh-think-translate/tar.gz/ba71a9bb88f52bc7bbf42225cfb69f7ef8d16900"
   ```

2. That plugin's lockfile entry is left in the "no integrity" half-written state after an interrupted update + manifest rollback:

   ```yaml
   dsh-think-translate@https://gh-proxy.com/...:
     resolution: {tarball: https://gh-proxy.com/...}
     version: 1.0.5
   ```

   The market log showed the trigger: `update: dsh-think-translate -> https://gh-proxy.com/.../33a55a40... exit=1 TIMEOUT` followed by `rolled back manifest residue of the failed run`. `restoreManifestDeps` restores `package.json` but leaves the lockfile as-is, so the resolution survives with `tarball:` and no `integrity:`.

3. Install *any* other plugin (e.g. `dsh-vision-router`):

   ```
   ERR_PNPM_MISSING_TARBALL_INTEGRITY: Cannot install package
   "dsh-think-translate@https://gh-proxy.com/...": its lockfile entry has no "integrity" field,
   so pnpm cannot verify the downloaded tarball.
   ```

   The UI shows the wrapper tail + raw stack. Nothing in `classifyPnpmFailure` (`src/pnpm-compat.ts`) matches, so no bilingual explanation and no automatic recovery is offered.

## Why this matters

It is a variant of the existing ghost-entry class (#65/#289): a corrupt lockfile entry blocks installs of *unrelated* packages. The tarball-integrity variant is specific to the accelerated `github:` → gh-proxy codeload-tarball route (the same path that makes large downloads timeout-prone in #83), so it will keep occurring whenever a download is interrupted mid-update.

## Suggested handling

- Add a `classifyPnpmFailure` case for `ERR_PNPM_MISSING_TARBALL_INTEGRITY` with an actionable bilingual message: name the package whose entry is missing integrity, and tell the user the lockfile entry needs its `integrity` field restored (re-fetch the tarball to compute sha512, or re-run `pnpm install` to reconcile).
- Consider automatic recovery where safe: for a tarball-URL dep whose download previously aborted, re-fetching the tarball and computing its `sha512` integrity is deterministic and non-destructive. (Plain `pnpm install` alone may keep failing because the stale entry is what pnpm refuses to verify.)

## Manual workaround (confirmed)

Download the tarball at the locked URL, compute `sha512` (SRI base64), and add `integrity: sha512-…` to the entry's `resolution`. After that the install succeeds.

Jstn-1g:
Closes #367.

## What changed
- classify `ERR_PNPM_MISSING_TARBALL_INTEGRITY` with a bilingual actionable message
- decode the actual `pnpm --reporter=ndjson` error payload before extracting the affected package
- safely name only canonical quoted `name@https-url` packages; use a generic fallback otherwise
- keep it non-recoverable and never generate integrity for unverified bytes automatically

## Verification
- [x] `npm test` — 969 passed, 1 skipped
- [x] `npm run check` — typecheck, build, and restart smoke pass
- [x] regression covers human output plus realistic escaped NDJSON, scoped/unscoped packages, and false-positive shapes
- [x] no version bump; no client source change

Exact head: `ee9815f439319acc72353430e4560a5b83ead4cb`

### DIFF
--- a/src/pnpm-compat.ts
+++ b/src/pnpm-compat.ts
@@ -45,7 +45,7 @@ export const HOST_NAMESPACE_RE = /^@deepseek-ai\//
 export interface PnpmFailure {
   code: 'adding-to-root' | 'not-a-workspace' | 'hoist-pattern-diff' | 'pnpm-missing' | 'release-age-violation'
     | 'ignored-builds' | 'git-prepare-not-allowed' | 'fetch-404' | 'transient-network' | 'fetch-timeout'
-    | 'unexpected-store' | 'patch-failed'
+    | 'unexpected-store' | 'patch-failed' | 'missing-tarball-integrity'
   /** Bilingual, actionable message shown to the user instead of the raw wall of text. */
   message: string
   /** True when re-running `pnpm install` in the profile is the documented recovery. */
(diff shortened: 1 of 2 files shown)


Jstn-1g:
Superseded by the maintainer carry in #373, merged as eda07d97f9a57a574066aa0d03a5e2cbe86c92af and shipped in v1.32.0. Thank you for preserving the #368 / @Jstn-1g attribution in the adopted commit; closing this duplicate PR.

fkysly:
已合并，随 **1.32.0** 发布。谢谢！

PR #374, opened 2026-08-27
fix: ignore direct profile shadows when the in-box host is hidden

Jstn-1g:
### What changed

Small follow-up to #371's Desktop preflight fix.

When Desktop cannot expose the DSH installation directory, an old profile-local copy of an official in-box bundle can shadow the host-maintained fallback at `<profiles>/node_modules`. The validator then either rejects a healthy update from the stale manifest or omits the real host layer and misses composition conflicts.

- Ignore only the direct `<profile>/node_modules/<official-bundle>` shadow when the host installation is hidden.
- Continue the remaining Node search paths, including Desktop's healed `<profiles>/node_modules` fallback.
- Preserve normal install-first resolution when the DSH installation is known.
- Leave community-bundle resolution unchanged.
- Cover direct-shadow-only, healed-parent-fallback, and duplicate-conflict behavior in both diagnostics and trial validation.

### Why this is separate from #371

#371 correctly made a truly unlocatable in-box bundle `unresolvedInbox` instead of a fatal error. This closes the remaining precedence gap without changing that design or the update transaction.

### Verification

- [x] New stale-shadow/healed-fallback regressions fail against base `2877a1b` and pass with this patch.
- [x] `npm test -- --run tests/check.spec.ts tests/trial.spec.ts` — 61 passed.
- [x] `npm test` — 979 passed, 1 skipped.
- [x] `npm run check` — typecheck, build, and restart smoke passed.
- [x] No version bump; no client source changes.

Follow-up to #371.

### DIFF
--- a/src/check.ts
+++ b/src/check.ts
@@ -379,15 +379,29 @@ function readNodeModulesVersion(base: string, name: string): string | null {
  * hoisting (`<profiles>/node_modules/…` when the profile lives under
  * `<profiles>/<name>`) and mirrors the Loader's package search roots.
  */
-function resolvePackageDir(anchorPackageJson: string, name: string): string | null {
+function resolvePackageDir(
+  anchorPackageJson: string,
+  name: string,
+  ignoredPackageDirectory?: string,
+): string | null {
   let paths: string[] = []
   try {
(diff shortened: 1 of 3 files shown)


fkysly:
已合并，随 **1.33.0** 发布。谢谢！

PR #375, opened 2026-08-27
feat: expose a stable plugin update API

duhu2000:
### Summary

- add a capability-discovered, versioned same-origin update API under `/dsh-market/api/v1`
- expose installed/latest versions plus asynchronous operation progress and stable failure codes
- expose operation-scoped compatibility rollback without leaking the Market's private rollback token
- delegate update, rollback, and restart to the existing guarded route handlers instead of duplicating package-manager logic
- document the v1 compatibility policy and export its operation types

### Why

Plugin-owned settings pages currently have to open the Market UI or couple themselves to private response fields. MCP Connector needs an in-page update action, but it should not spawn npm/pnpm or reimplement Market's compatibility, rollback, and lifecycle safeguards.

This contract keeps the Market as the single update authority while allowing other plugins to provide a small, capability-gated update surface.

### Safety and lifecycle

- mutation calls retain the existing same-origin protection
- restart still uses the existing loopback, forwarding-header, supervisor, and busy guards
- only one public update operation can run at a time
- operation IDs are boot-scoped and at most 50 records are retained
- Desktop/supervised hosts advertise restart as unsupported so callers can hand lifecycle control back to the host
- clients without v1 discovery are expected to fall back to opening the Market

### Verification

- [x] `npm run check`
- [x] `npm test` — 48 test files, 987 tests passed
- [x] new route tests cover capabilities, update checks, async progress, concurrent rejection, agent guards, rollback, and restart delegation
- [x] no package version bump
- [x] no `src/client/` changes; generated client output is unchanged


fkysly:
已合并，随 **1.33.0** 发布。谢谢！

PR #376, opened 2026-08-27
feat: see what changed before you take an update

fkysly:
承 #366（@00080000），实现他自己提的 #294。

**运行时零 GitHub 调用**——说明由目录侧每天探一次、发成 `dsh-plugin-updates` 包，全体用户共享。这点很关键：匿名 REST 配额是每出口 IP 每小时 60 次共享，正是 #349 里把更新检查整个打死的那个东西。

与先落地的 #347 做了调和：他们的简介块去掉（用户备注已经取代它），他们的 `notes` 状态改名 `updateNotes`——我那边把 `notes` 用作每插件备注表了，两者不是一回事。

**已用今天发布的包端到端验证**：从 npm 载入 2225 条、repo key 命中、取回真实 release 正文。

987 测试全过。


fkysly:
Landed as #366 (@00080000's PR, implementing their own #294) — same commit, with them recorded as co-author. Closing this one so the work has a single home.

Rebased onto 1.34.0 on the way in: the update route moved under `captureLegacy` on main and the comments feature appended to the same stylesheet, so both were merged by hand and the client bundle rebuilt from source rather than taken from either side. `npm run check` and 1025 tests pass; the `/dsh-market/changelog` route was checked against a real dsh — real npm publish dates come back for an out-of-date plugin, a missing `name` is a 400, and a plugin with nothing to say degrades to `kind: "none"` rather than an error.

PR #383, opened 2026-08-27
fix(update): reject resolved version downgrades

duhu2000:
## What changed

- Verify the npm package version that pnpm actually installed before an update is reported as successful.
- Reject and automatically restore the pre-update build when pnpm resolves an older version or a version different from the registry target.
- Preserve `DOWNGRADE_DETECTED` and `RESOLVED_VERSION_MISMATCH` through the public Update API v1 operation model.
- Document the integrity guarantees and the client-side defense-in-depth requirement.

## Why

With pnpm `minimumReleaseAge` (or a lagging mirror), `pnpm add package@latest` can exit 0 while resolving an older release. The previous stale-update check only rejected an unchanged version, so a real downgrade such as 0.2.24 -> 0.2.23 was treated as a successful update.

## Checklist

- [x] `npm test` passes locally (50 test files, 1013 tests)
- [x] `npm run check` passes locally
- [x] Regression tests fail without the fix and cover both downgrade and unexpected-newer-version resolution
- [x] No `src/client/` changes; generated client bundle is unchanged

## Verification

- TypeScript type checks passed
- Server and client builds passed
- Restart smoke test passed
- Public Update API v1 failure-code normalization tests passed

### DIFF
--- a/UPDATE-API-V1.md
+++ b/UPDATE-API-V1.md
@@ -74,6 +74,22 @@ pnpm provides it. Terminal operations include:
 - a stable failure code, bounded user-facing message and retryability;
 - whether a compatibility rollback is currently available.
 
+For npm updates, Market verifies the package version that pnpm actually placed
+on disk before reporting success. This prevents pnpm's release-age policy or a
+lagging registry mirror from silently turning `@latest` into an older or
+unexpected build. A mismatch is restored from the pre-update recovery point
+and reported as one of these stable failures:
+
+- `DOWNGRADE_DETECTED`: the resolved version is older than the version present
+  before the operation; not retryable without a new target.
+- `RESOLVED_VERSION_MISMATCH`: the resolved version differs from the registry
+  target checked immediately before installation; retryable after registry or
+  mirror convergence.
+
+Provider clients should still compare `beforeVersion`, the target they showed
+to the user, and `installedVersion` before offering restart. That independent
+check protects clients connected to another compatible provider implementation.
+
 Up to 50 operation records live in the current Host process. A boot id is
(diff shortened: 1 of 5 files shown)


fkysly:
Good catch, and the right layer to catch it at — pnpm exiting 0 on a resolution it never promised is exactly the kind of thing that has to be checked against what landed on disk rather than against what we asked for.

I pushed one commit to your branch. The exact-equality comparison also rejects the case where the target moves *forward*:

```
registry latest = 1.2.0   →   pnpm starts downloading
                              author publishes 1.2.1
pnpm resolves 1.2.1       →   afterVersion !== expectedNpmVersion  →  rolled back
```

A large plugin leaves minutes of window for that, and the result is a good, newer build reported as a failure — the mirror image of the bug this guard exists to prevent. Confirmed against your branch before changing anything:

```
FAIL  accepts a release published while the install was still running
AssertionError: expected { ok: false, … } to match object { ok: true }
```

The comparison is now ordered rather than exact: below the target is a mismatch, above it is accepted. Both of your cases still fail as they should — the 1.1.0-for-1.2.0 test resolves *below* the target, so the narrowing does not touch it.

Two smaller things that came out of it:

- `compareVersions` returns `number | null`, null when either side is not parseable semver. Where it cannot compare, the code falls back to your exact check — with no way to tell forward from back, the stricter behaviour is the right default.
- `afterVersion === null` is still treated as a mismatch. If nothing is readable on disk after a "successful" install, restoring is right even if `RESOLVED_VERSION_MISMATCH` is a slightly generous name for it.

`npm run check` and 1014 tests pass. Merging once CI is green.

PR #409, opened 2026-08-29
feat: rank market search by field relevance

duhu2000:
## Problem

Discover search currently acts as a boolean filter and then applies downloads/stars/date ordering to every match. A very popular plugin that only mentions the query in its description can therefore rank ahead of a package whose name directly matches the query.

## Change

- normalize case and punctuation so package names and human queries match consistently
- search both catalog names and npm package names
- keep every localized description searchable, with the active locale weighted first
- rank package identity matches above owner, description, and category matches
- preserve the selected downloads/stars/date order as the tie-breaker
- preserve existing behavior when the search box is empty

This is field-aware and generic; it contains no package-specific boosts.

## Validation

- `npx vitest run tests/market-data.spec.ts` — 43 passed
- `npm test` — 52 files / 1045 tests passed
- `npm run test:compat` — 11 passed
- `npm run check` — typecheck, build, and restart smoke passed
- current registry snapshot: `mcp` puts package-name matches before description-only matches while keeping downloads as the tie-breaker


fkysly (changes requested):
The ranking itself is right, and the results are clearly better — I checked against the live 2,452-entry snapshot and `dsh-loop` goes from 2 matches to 30, `task runner` from 0 to 1, because normalizing punctuation finally lets a hyphenated package name match the way people type it. I want this. One thing to fix first.

## It costs 20× on every keystroke

Benchmarked both implementations against `data/registry-snapshot.json` (2,452 plugins), 20 calls each, warm:

| query | main | this PR | results (main → PR) |
|---|---|---|---|
| `""` | 0.8 ms | 3.9 ms | 2451 → 2451 |
| `mcp` | 1.1 ms | 20.9 ms | 120 → 121 |
| `memory` | 1.0 ms | 19.9 ms | 123 → 170 |
| `记忆` | 1.0 ms | 19.3 ms | 151 → 151 |
| `dsh-loop` | 1.0 ms | 19.7 ms | 2 → 30 |
| `task runner` | 1.0 ms | 19.8 ms | 0 → 1 |

That is on an M-series Mac. `visiblePlugins` runs on every keystroke, so a modest Windows laptop lands somewhere around 60–100 ms of blocking JS per character — into the range where typing feels like it is dragging.

The cause is not the ranking; it is that `searchText()` re-normalizes the *same* field strings every call. NFKC normalize plus a unicode-property regex, over roughly eight fields × 2,452 plugins ≈ 20,000 strings, per keystroke, to produce byte-identical output each time.

## The fix is a cache, and I verified it

I applied this to your branch and re-ran the same benchmark:

```ts
const searchTextCache = new Map<string, string>()
function searchText(value: string): string {
  const hit = searchTextCache.get(value)
  if (hit !== undefined) return hit
  const out = value.normalize('NFKC').toLowerCase().replace(/[^\p{L}\p{N}]+/gu, ' ').trim().replace(/\s+/g, ' ')
  searchTextCache.set(value, out)
  return out
}
```

| query | before | after |
|---|---|---|
| `mcp` | 20.9 ms | 2.1 ms |
| `memory` | 19.9 ms | 2.1 ms |
| `dsh-loop` | 19.7 ms | 2.1 ms |
| `task runner` | 19.8 ms | 2.0 ms |

Same results, 10× back. Feel free to take that verbatim.

One caveat worth handling however you prefer: a plain `Map` also caches every distinct *query* string, which grows by one entry per keystroke for the life of the page. Bounded by the catalog for field values, unbounded in principle for queries. Either skip the cache for the query (it is normalized once per call anyway, so it gains nothing), or key a `WeakMap` off the plugin object so the whole thing is collected with the catalog.

## Two smaller notes, neither blocking

**A punctuation-only query now shows everything.** `searchText('---')` is `''`, which takes the `query === ''` path, so the full list comes back where it used to return nothing. Arguably an improvement — that is not a search — just flagging it as an intended consequence rather than an accident.

**Multi-word queries got more permissive**, and I think correctly: `task runner` used to need that literal phrase inside one field, and now needs both words in the same field. That is why it goes 0 → 1 rather than staying at 0. Worth a line in the tests so the intent is recorded, since it is a behaviour change and not just a ranking change.

duhu2000:
Addressed the performance review in `aaa9d1e` and synchronized current `main` in merge commit `36107d2`:

- catalog fields are normalized through a `WeakMap<RegistryPlugin, Map<string, string>>`, so repeated keystrokes reuse per-entry values and replaced catalogs remain collectible
- the raw query is intentionally normalized once per call and is not cached
- tests now lock in punctuation-only input as no search and require all multi-word tokens to match within one field
- regenerated the client bundle after syncing the seven upstream commits

Local verification on the combined tree: `npm run check`, 53 test files / 1064 tests, and 11 compatibility tests all pass. The branch conflict is resolved and fresh CI is running. Please re-review when convenient.

duhu2000:
@fkysly Latest head e5f26ef is now MERGEABLE/CLEAN against current main. Revalidated after the upstream sync: the focused market-data suite passes 44/44, npm run check passes, and the full suite passes 55/55 files and 1127/1127 tests. All six required GitHub checks are green, including pnpm-compat and Windows Web E2E. The requested WeakMap cache and the punctuation/multi-word behavior tests remain in aaa9d1e. Could you please re-review when convenient?

fkysly (approved):
重测过了,阻塞的那条解决了。收。

在**真实目录快照(2663 条)**上重跑,和 main 对照:

| | main | 本 PR |
|---|---|---|
| 稳定态每次按键 | 0.83 ms | **1.93 ms** |
| 第二次按键 | 2.02 ms | 3.15 ms |
| `dsh-loop` 命中 | 2 | **36** |
| `task runner` 命中 | 0 | **2** |

之前那次 20 倍的回归(1 ms → 20 ms)没有了。剩下约 2.3 倍的差距是排序本身的成本,两边都远在一帧以内,打字不会拖。

**排序质量我也单独看了**,不是「结果多了」而已:

```
"dsh-loop" → 36 条,前 5:  dsh-loop / dsh-loop-engine / dsh-outcome-loop / dsh-review-loop / dsh-frostfin
"task runner" → 2 条:     dsh-task-runner / dsh-lowtide#dsh
"theme" → 128 条,前 5:    …-theme / dsh-theme-endfield / dsh-kimino-theme / …
```

精确匹配在第一位,近似变体紧随,松散匹配排在后面 —— 多出来的结果不会挤掉用户真正想找的那个。这正是「更多结果」值得要的前提。

## 一个非阻塞的观察

稳定态的数字掩盖了一件事:**第一次按键要 26.84 ms**(main 是 1.82 ms)。那是 WeakMap 空的时候,一次性把 2663 条的字段全部 NFKC 规范化的代价。

不阻塞合并 —— 它是每个会话一次,而且比原来「每次按键 20 ms」好得多。但我在这台 M 系列 Mac 上测到 27 ms,普通 Windows 笔记本上按经验会到 80–150 ms,那就是用户在**第一个字符**上能感觉到的一下卡顿。

如果你想再进一步,建议是:目录加载完成后用 `requestIdleCallback` 预热一次缓存,把这笔账在用户开始打字之前付掉。大概几行,而且没有正确性风险(缓存内容和按需算出来的完全一样)。可以另开 PR,也可以不做。

55 files / 1127 passed,`npm run check` 通过,六项 CI 全绿。

谢谢你把 WeakMap 那版和标点/多词的行为测试都补齐了,也谢谢耐心 —— 这个改动的收益是实打实的,我只是不想让它以每次按键变慢 20 倍为代价。

Issue #411, opened 2026-08-29
功能规划：Agent 安全排空后自动执行插件变更 / Agent-safe drain mode


Harzva:
### 你遇到的问题 / The problem

Since #199, install, update, and uninstall correctly refuse to mutate plugin
files while any Agent is running. The safety boundary is sound, but the user
experience is a dead end: the operation becomes a failed task, the user has to
find every running session, wait or cancel it, then return and retry manually.

This is especially awkward for long-running or background Agents. Cancelling
them is not equivalent to pausing them: an active model stream or tool call
cannot be resumed from an arbitrary instruction boundary, and retrying it may
repeat side effects.

The desired behavior is to preserve the #199 invariant while removing the
manual retry loop.

### 你设想的做法 / What you have in mind

Add an Agent-safe **drain mode / 排空模式** for deferred plugin mutations:

1. An install/update/uninstall requested while Agents are running becomes a
   Host-owned `draining` operation instead of a terminal 409 failure.
2. Existing Agent turns finish naturally; they are never cancelled.
3. When an Agent reaches true idle, the coordinator claims its public
   `runMaintenance()` boundary. New inbox work remains queued behind that
   maintenance task, while public Agent status stays idle.
4. Once every live Agent is idle and maintenance-held, the Market acquires its
   existing mutation lock, rechecks the invariant, and runs the package change.
5. After validation settles, the maintenance tasks are released. DSH's
   existing wake-latch behavior then continues queued Agent work automatically.

This is safe-boundary continuation, not arbitrary process suspension. On hosts
without `Agent.whenIdle()` + `Agent.runMaintenance()`, the current 409 behavior
should remain the compatibility fallback; a polling-only implementation leaves
a race where a new turn can start between the idle check and pnpm.

Suggested v1 scope:

- one Host-owned deferred mutation at a time, preserving the current single
  mutation lane;
- install/update/uninstall plus the public update API projection;
- process-local operation retention (survives closing Settings, not Host
  restart), matching the current public update API;
- a cancellable `draining` task row showing the number of Agents still active;
- no force bypass, no automatic Agent cancellation, and no automatic Host
  restart;
- immutable target identity captured at enqueue time so a delayed operation
  cannot silently install a different release.

I am preparing a specification-only PR with the state machine, API/UX contract,
failure behavior, race invariants, and acceptance tests. Runtime implementation
would start only after the scope is accepted.

Related: #162 (operation queue UX), #186/#199 (running-Agent safety gate), and
[...]

Issue #320, opened 2026-08-24
Lifecycle webhook/event seam: let plugins observe and gate marketplace installs


GavinTomlins:
### What happens today

Installing from the market is a one-click flow that ends in pnpm fetching and mounting an
unvetted npm package **in the harness's own process** — same memory, same filesystem sandbox,
same LLM keys. The only pre-install security gate that exists today is my plugin,
[dsh-security-scanner](https://github.com/GavinTomlins/dsh-security-scanner): it intercepts the
CLI `dsh plugin add` path before pnpm runs, scans every published file of the target version
with an LLM (snickers methodology: prompt injection, tool poisoning, malware payloads, toxic
flows), and returns one of three verdicts — `clear`, `flagged`, `unavailable` — failing closed
on anything that isn't `clear`.

On the market path, that gate is enforced only **incidentally**: dshmarket shells out to the
installer as a child process with closed stdin, so a flagged/unavailable spec gets
auto-rejected because there is no human to answer `y/N`. It works, but nothing about it is
contractual.

### Where it breaks

1. **Governance by accident.** If dshmarket changes *how* it invokes the installer (in-process
   call, different spawn flags, queueing), the scanner's enforcement silently disappears —
   there is no defined point where a scanner participates in a market install.
2. **No scan UX in the market UI.** While a scan runs (~5–15 s per spec), the market card just
   spins. There is no way to stream "scanning → ✓ clear / ✗ flagged (6 findings)" into the
   flow the user actually clicked, and no way to link the market card to the scan record.
3. **No say for other extensions.** Nothing else can observe, veto, annotate, or enrich an
   install. Security is the motivating case, but not the only one.
4. **Installs always land in the running profile.** There is no seam for a consumer to choose
   the target profile per plugin (e.g. themes → the `dsh-web-app` profile, headless tools → the
   base profile). [#129](https://github.com/dsh-market/dsh-market/issues/129) shows the
   resulting confusion when a package lands in a profile bundle layer the user didn't expect.
5. **Nothing leaves the process.** Org policy engines, CI checks, and chat notifications have
   no machine-readable signal that an install was requested/succeeded/failed.

### Feature set

Emit lifecycle events at four defined points in the market's install flow:

| Event | Kind | Payload |
|---|---|---|
| `marketplace/install-requested` | **blocking** (listeners may deny) | spec string, resolved `name@version`, source listing id, requested-by context, mutable `targetProfile` |
[...]

PR #412, opened 2026-08-29
docs: specify Agent-safe drain mode for deferred mutations

Harzva:
## Summary

- specify an Agent-safe drain mode for deferred install/update/uninstall operations
- let active turns finish naturally, then hold idle Agents through DSH's public `runMaintenance()` boundary
- keep new Agent work in the existing inbox and release it automatically after mutation and validation
- preserve current 409 behavior as the compatibility fallback on Hosts without drain primitives

## Why

#199 correctly prevents plugin files from changing under a running Agent, but the current 409 turns that safety boundary into a manual stop/wait/retry loop. Cancelling is not suspension, and polling for idle leaves a race before pnpm starts.

This proposal uses `Agent.whenIdle()` plus `Agent.runMaintenance()` as a cooperative admission barrier. It does not interrupt a model stream or tool call, synthesize prompts, or claim arbitrary execution can be resumed.

## What the specification locks down

- goals, non-goals, and a deliberately bounded v1 scope
- first-party UI behavior and Host-owned operation recovery after page navigation
- internal and public update API state projections
- immutable delayed-operation intent
- race-free Agent creation/admission handling
- cancellation, unload, failure, and restart behavior
- privacy limits, compatibility fallback, rollout, and acceptance tests

This is a specification-only PR. Runtime implementation should begin after the product and lifecycle boundaries are accepted.

Tracks #411. Related to #162, #186/#199, and #320.

## Validation

- `git diff --check`
- staged-content scan for credentials, environment values, raw logs, and machine-private paths
- documentation-only change; no generated artifacts or runtime code changed

### DIFF
--- a/DRAIN-MODE-SPEC.md
+++ b/DRAIN-MODE-SPEC.md
@@ -1,0 +1,459 @@
+# Agent-safe drain mode / Agent 安全排空模式
+
+> Status: proposal
+>
+> Tracking: [#411](https://github.com/dsh-market/dsh-market/issues/411)
+>
+> Related: [#162](https://github.com/dsh-market/dsh-market/issues/162),
+> [#186](https://github.com/dsh-market/dsh-market/issues/186),
+> [#199](https://github.com/dsh-market/dsh-market/pull/199), and
+> [#320](https://github.com/dsh-market/dsh-market/issues/320)
+
+## Summary
+
+When an Agent is running, dshmarket currently rejects install, update, and
+uninstall requests. That guard prevents a live Agent from mixing already-loaded
+code with plugin files replaced underneath it, but it leaves the user with a
+manual recovery loop: find every active Agent, wait or cancel, return to the
+Market, and retry.
+
+Drain mode keeps the same safety invariant and replaces the dead end with a
+deferred operation. Existing Agent turns finish naturally. As each Agent
+reaches true idle, dshmarket holds it at DSH's maintenance boundary, where new
+inbox work remains queued. Once every live Agent is held, dshmarket performs the
+mutation, validates it, releases the holds, and lets queued Agent work continue.
+
+This is cooperative continuation at a defined runtime boundary. It is not
+arbitrary suspension of a model stream, tool call, or operating-system process.
+
+## Context and invariant
+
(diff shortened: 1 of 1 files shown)


fkysly:
Closing the PR, keeping the work — and I want to be clear that the second part is the real one.

**You checked the host before proposing.** I verified the two primitives against `@deepseek-ai/dsh-agent/lib/types/runtime-types.d.ts` in the installed host, and they are exactly what you describe:

> `runMaintenance<T>(task)` — "Run one non-turn maintenance task from the true idle phase. The task starts synchronously after claiming that phase; later waking input remains in the inbox until the task settles, while public status stays `idle`." … "**throws synchronously when turn-driving or another maintenance task already owns the agent**"

That last clause is the whole argument, and you made it yourself in the issue: a polling implementation leaves a gap between "observed idle" and "pnpm started" where a new turn can open. `runMaintenance` claims the idle phase atomically, so the gap does not exist. That is a genuinely better answer than the one I would have reached for.

## Why not merge the file

A 459-line `DRAIN-MODE-SPEC.md` at the repository root is the part I am declining, not the design. This codebase keeps its rationale in code comments and commit messages — deliberately, because that is the copy that gets read when someone is changing the line and cannot go stale independently of it. `docs/` is the published website, not a design area. An unimplemented spec in-tree becomes wrong the first time the design moves, and nothing forces it to be updated.

So it is recorded where unformed work belongs here, with your reasoning and the API evidence intact.

## What I think has to be settled before anyone implements

The mechanism is not what worries me — the product semantics are:

1. **Who owns a deferred mutation once the user navigates away?** Your FR-10 covers process restart, but "I clicked install, closed the tab, and something changed my plugins twenty minutes later" needs an answer a user would accept.
2. **How does a drained operation fail loudly enough?** The 409 is a dead end, but it is a dead end the user is *looking at*. A deferred failure has to find someone.
3. **How does it coexist with the existing banners?** There are already refresh and restart banners, and merging them was itself a fix — the report was "为啥有三个状态横幅啊，太奇怪了". A third lifecycle needs to land inside that, not beside it.

Answer those and the implementation is the easy half. Thanks for doing this properly — the issue stays open.

Issue #213, opened 2026-08-19
[Bug] External removal + refresh can wedge Web UI when a standalone plugin overlaps an aggregate bundle


wisdomqin:
## Summary / 摘要

After removing a standalone plugin that overlaps with a plugin already supplied by an aggregate bundle, the current page can remain usable, but a manual page refresh wedges the whole Web UI with a client-loader error.

This looks related to #37, but it is a different variant:

- #37 was a removed bundle whose client URL became 404.
- In this case the failing client bundle is a **different transitive plugin**, its URL still returns **HTTP 200**, and its content hash matches the requested `rev`.

## Environment / 环境

- macOS arm64
- DSH Desktop: 2.0.1
- `@deepseek-ai/dsh`: `0.1.0-rc.7`
- dshmarket: `1.14.1`
- Aggregate bundle: `@linxin666/dsh-web-ui-all@0.2.2`
- Standalone overlapping plugin: `dsh-better-sidebar@0.13.0`

The current upstream `main` / v1.15.0 still appears to use the same conditional cleanup path described below.

## Profile before removal

```json
{
  "dependencies": {
    "@linxin666/dsh-web-ui-all": "^0.2.2",
    "dsh-better-sidebar": "^0.13.0",
    "dsh-dream-skin": "^0.3.0",
    "dshmarket": "1.13.1"
  },
  "dsh": {
    "profile": {
      "bundles": [
        "@deepseek-ai/dsh-base",
        "@deepseek-ai/dsh-web-app",
        "dsh-better-sidebar",
        "dsh-dream-skin",
        "dshmarket",
        "@linxin666/dsh-web-ui-all"
      ]
    }
  }
}
```

`@linxin666/dsh-web-ui-all` itself depends on and inserts both:

- `dsh-better-sidebar`
- `@linxin666/dsh-client-ui-web-ui-settings`

## Steps to reproduce / 复现步骤

1. Install `@linxin666/dsh-web-ui-all@0.2.2`.
2. Also install `dsh-better-sidebar@0.13.0` as a standalone profile bundle.
3. On the next boot, the two activation sources conflict on `/sidebar/api`.
4. Remove the standalone `dsh-better-sidebar` dependency and bundle entry. In the observed case this was performed by DSH Desktop startup recovery, i.e. an external profile mutation rather than the market's own uninstall button.
5. Launch Harness. The page can initially open.
6. Refresh the page.
7. The shell fails with:

```
[...]

PR #415, opened 2026-08-29
fix: uninstall client-part plugin refresh prompt with hot gate

fenglin-dev:
## 修复内容

本 PR 修复卸载带有 client 部分的插件后不提示刷新页面的问题。

### 问题

卸载带有 client 部分（`dsh.client` in package.json）的插件后，已注入到页面的 UI bundle 仍然在运行，直到用户手动刷新页面。而启用/关闭插件时会正确显示刷新提示横幅，卸载时却没有。

### 修复

**后端（src/routes.ts）：**

1. **提前捕获 `hadClientPart`**：在 `runPlugin(['remove', name])` 之前调用 `packageHasClientPart()`，因为卸载成功后 package.json 可能已被删除，事后检查会永远返回 false。

2. **添加 hot 门控**：`refresh: ok && hot && hadClientPart`
   - `hot=true` + 有 client 部分 → 显示 refresh 横幅（没有 restart 横幅）
   - `hot=false` + 有 client 部分 → 不显示 refresh 横幅（已有 restart 横幅，避免双重横幅）
   - 没有 client 部分 → 不显示横幅

3. **删除了 install 路由中错误的 refresh 块**：install 处理器中没有 `name` 变量，这个块是从 uninstall 误复制的。

**前端（src/client/MarketSection.tsx）：**

在 `doUninstall` 函数中处理 `body.refresh` 字段，当为 true 时将插件加入待刷新列表（显示与启用/关闭相同的刷新提示横幅），而不是清除它。

### 修改文件

- `src/routes.ts`
- `src/client/MarketSection.tsx`

### 致谢

感谢 @fkysly 的详细审查反馈。


Jstn-1g (changes requested):
Thanks for tackling both reports. I reviewed exact head `9e3bf343607310332d1355e06161c5357e4189fa` and reproduced the lockfile failure path. I think this needs changes before merge:

1. **The branch currently cannot build.** In `src/dsh-cli.ts`, the two new `logEvent` messages at lines 802 and 806 are bare template text rather than strings/template literals. In `src/routes.ts:3317`, the install response also references `name`, which is not defined in that handler; this refresh block appears to have been copied from uninstall into install.

2. **The lockfile repair is not failure-atomic.** `cleanupPnpmLockForUrlTarball()` deletes the profile's entire `pnpm-lock.yaml` before the CLI process starts and before the Desktop host accepts the operation. In a focused Desktop test, I started with a real lockfile and made the host reject as busy; the adapter returned `exitCode: 127` / `busy: true`, but the lockfile was already gone. `restoreProfileManifest()` only restores `package.json`, so refusal, spawn error, timeout, cancellation, or later validation failure can discard integrity/resolution pins for unrelated dependencies. Please avoid whole-file preflight deletion. Prefer a verified entry-scoped repair, or preserve the exact file and restore it on every non-committed exit inside the authoritative package-runtime transaction.

3. **The URL matcher misses the production form cited in the PR.** `https://codeload.github.com/<owner>/<repo>/tar.gz/<sha>` does not match `\.(tar\.gz|tgz)(\?|$|#)`, because `/` follows `tar.gz`.

4. **The uninstall refresh signal is computed too late.** `packageHasClientPart()` reads `node_modules/<name>/package.json`, but it is called after `runPlugin(['remove', name])`; on a normal successful uninstall that metadata is gone, so `refresh: true` cannot be produced. The `ok` gate also excludes a reconciled half-uninstall whose package is gone. Capture `hadClientPart` before mutation and derive refresh from the final removed state. Please also clear any stale hot marker when adding the name to the refresh set.

Please add regressions for CLI spawn failure, Desktop busy/refusal/timeout/cancel, byte-identical lockfile preservation, the actual `/tar.gz/<sha>` URL shape, successful and reconciled client-package uninstall refresh, and client hot/refresh state. There are no CI checks attached to the PR right now, so those tests plus the normal typecheck/build suite are important before this is safe to merge.

fkysly (changes requested):
Thanks — the second fix is a real bug that nobody had noticed, and I want it. The first one I have to turn down on substance, and there are two mechanical problems that stop the branch building at all. Details so this is actionable rather than just a no.

## Blocking: the branch does not compile

Both new `logEvent` calls lost their backticks:

```js
logEvent('info', 'install', url-tarball target ${target}: removed stale pnpm-lock.yaml (no integrity field))
```

That is a syntax error, so `npm run check` cannot have passed on what was pushed — whatever you tested locally is not what is on the branch. (No CI ran here either; fork PRs need a maintainer to approve the run, which is why it looks quiet rather than red.)

## Blocking: a UTF-8 BOM was added to three files

`src/dsh-cli.ts`, `src/routes.ts` and `src/client/MarketSection.tsx` all gained a leading `EF BB BF`:

```
$ git show <branch>:src/dsh-cli.ts | head -c 3 | xxd -p
efbbbf
```

That is why each diff opens with `-/**` / `+﻿/**` on an unchanged line. Editor default on Windows, usually — worth turning off, since it silently touches every file you save.

## Fix 1: I don't want to delete the user's lockfile

Two reasons, and the second is the bigger one.

**It is more destructive than the problem.** `pnpm-lock.yaml` pins the whole profile, not just the plugin being installed. Deleting it makes the next install re-resolve every other plugin's dependency tree, so unrelated, working plugins can move versions because someone installed a tarball. That is the same shape as the bug 1.35.0 shipped to fix, where one install left a profile that would not boot — a plugin operation must not be able to change plugins it was not asked to touch.

**And `ERR_PNPM_MISSING_TARBALL_INTEGRITY` was fixed at its source yesterday**, in #392 (released in 1.36.0). The cause was narrower than "URL tarballs carry no integrity": when the China mirror handed pnpm a prefix-proxied `https://proxy/https://codeload.github.com/…` URL, the codeload hostname was hidden, so pnpm stopped treating it as gitHosted, reused store bytes from an earlier resolution, and wrote a lockfile entry with no integrity — which then failed closed forever. #392 stops producing that URL: the mirror now only resolves HEAD, and pnpm gets a canonical `github:owner/repo#<sha>`.

So please check whether you still reproduce this on 1.36.0. If you do, I want that case — with your `pnpm-lock.yaml` entry and the exact install target — because it would be a shape #392 does not cover, and the right fix would be to stop generating the bad entry, or to repair that one entry, rather than to discard the file.

If you are stuck on a lockfile poisoned *before* 1.36.0, deleting it by hand once is a fine one-time repair. It is being automatic, on every tarball install, that I am objecting to.

## Fix 2: right bug, wrong condition — it would double-banner

You are correct that this is a gap, and the diagnosis is exact: `hot` means the *host-side* activation was torn down, and nothing removes the already-injected browser bundle, so a client-part plugin keeps rendering after it is uninstalled. `clearPendingRefresh(name)` is actively wrong there.

But `refresh: ok && packageHasClientPart(...)` is unconditional, and the non-hot path already prompts:

- `hot === false` → the client does `setRemovedCount(n => n + 1)`, which feeds `pendingRestart` and shows the **restart** banner. A restart reloads the page, so adding a refresh banner gives two banners for one uninstall — the exact complaint the merge in `pendingRefreshNames` was written to fix ("为啥有三个状态横幅啊，太奇怪了").
- `hot === true` → no restart banner, no refresh banner, and the UI is still on screen. **That is the whole bug.**

So gate it on `hot`:

```ts
refresh: ok && hot && packageHasClientPart(activeProfileDir, name),
```

and the client change can stay as you wrote it.

A regression test would need to cover both arms — hot uninstall of a client-part plugin shows exactly one banner, non-hot shows exactly one banner, and neither shows two.

## What I'd like

Split fix 2 out into its own PR with the `hot` gate, the BOM removed, and the backticks restored. I'll merge that. Fix 1 stays open as a question: does it still reproduce on 1.36.0?

fenglin-dev:
Thanks for the detailed review! I've addressed all the feedback:

**1. Reverted Fix 1 (lockfile deletion)**
- The root cause (ERR_PNPM_MISSING_TARBALL_INTEGRITY) was already fixed in #392 (released in 1.36.0). Auto-deleting pnpm-lock.yaml is too destructive — it re-resolves every plugin's dependency tree and can break unrelated, working plugins.
- dsh-cli.ts is now back to original.

**2. Fixed Fix 2 (uninstall refresh) with hot gate**
- Capture `hadClientPart` BEFORE `runPlugin`: after a successful uninstall the package may be gone from node_modules, so a post-hoc check would always return false.
- Gate refresh on `hot`: non-hot uninstalls already show the restart banner, and adding a refresh banner there would double-banner (#213's pendingRefreshNames merge exists specifically to avoid that).
- Now: `refresh: ok && hot && hadClientPart`

**3. Removed erroneous refresh block from install route**
- The install handler had no `name` variable in scope — this block was copied from uninstall by mistake. Installs don't need a refresh prompt.

**4. Removed UTF-8 BOM from all modified files**
- src/routes.ts and src/client/MarketSection.tsx had leading EF BB BF (Windows editor default). This caused every diff to open with -/** / +/** on an unchanged line.

**5. Client-side change preserved**
- When `body.refresh === true`, add the plugin to the refresh set instead of clearing pending refresh.

Regarding regression tests: I wasn't able to run the test suite in this environment (no node_modules installed), but the changes are minimal and targeted. The logic is:
- hot=true + hadClientPart=true → refresh=true (shows refresh banner, no restart banner)
- hot=false + hadClientPart=true → refresh=false (shows restart banner only)
- hadClientPart=false → refresh=false (no banner needed)

Please let me know if there's anything else I should adjust!

fenglin-dev:
## 修复总结（更新）

已按照 @fkysly 的审查要求完成所有修改：

### 1. Fix 1 已完全撤销
- `src/dsh-cli.ts` 已恢复到 main 分支原始版本，不再包含 lockfile 删除逻辑
- 原因：ERR_PNPM_MISSING_TARBALL_INTEGRITY 的根本原因已在 #392 中修复（1.36.0 发布），自动删除 lockfile 太具破坏性

### 2. Fix 2 已修正（卸载 refresh 提示）
- **提前捕获 `hadClientPart`**：在 `runPlugin(['remove', name])` 之前调用 `packageHasClientPart()`，避免卸载成功后 package.json 被删除导致检测失败
- **添加 hot 门控**：`refresh: ok && hot && hadClientPart`
  - `hot=true` + 有 client 部分 → 显示 refresh 横幅（没有 restart 横幅）
  - `hot=false` + 有 client 部分 → 不显示 refresh 横幅（已有 restart 横幅，避免双重横幅）
  - 没有 client 部分 → 不显示横幅

### 3. 删除了 install 路由中错误的 refresh 块
- install 处理器中没有 `name` 变量，这个块是从 uninstall 误复制的

### 4. 移除了所有文件的 UTF-8 BOM
- `src/routes.ts` 和 `src/client/MarketSection.tsx` 的前导 `EF BB BF` 已移除

### 5. PR 标题和描述已更新
- 标题：`fix: uninstall client-part plugin refresh prompt with hot gate`
- 描述：已更新，只包含 Fix 2 的详细说明

### 修改文件
- `src/routes.ts`
- `src/client/MarketSection.tsx`

现在 PR 只包含 Fix 2，符合审查要求。请再次审查，谢谢！

fkysly:
Merging this. Two things worth saying, and the first is that **you corrected my review.**

## You caught a mistake in what I asked for

I wrote `refresh: ok && hot && packageHasClientPart(activeProfileDir, name)`. That is wrong: by the time it runs the package is gone from `node_modules`, so `packageHasClientPart` reads a file that no longer exists and returns `false` on every successful uninstall — the exact case the fix is for. Capturing `hadClientPart` **before** `runPlugin` is the correct shape and I had not thought it through. You also found and removed the stray `refresh` block I had not noticed was in the install route, where `name` is not even in scope.

## Writing the tests I asked for exposed a third arm, and it broke #340

The two arms hold: hot uninstall of a loaded plugin → exactly one banner (refresh, no restart); non-hot → exactly one (restart, no refresh). I verified they discriminate — reverting the client change fails the first one.

But there is a case neither of us considered. **Install a client-part plugin and uninstall it again without reloading.** The route answers `refresh: true`, correctly, because the package did declare `dsh.client`. Except that bundle was never injected into this page: the banner had been asking the user to reload *in order to get it*. Undoing the install nets to zero, so the banner has to go — and with this change it stayed.

That is #340 verbatim ("it was reporting session history, not pending work"), and the new test reproduced it:

```
× still stops asking when the undone plugin has a client part
  AssertionError: expected [ …(1) ] to have a length of +0 but got 1
```

**The server cannot make this call.** `refresh` says the package *had* a client part, not that this page ever loaded it. The client can: a name already in `hotNames` or `refreshNames` was installed during this session and was never live here. So the reload is requested only when it is not.

I pushed that plus the three tests onto your branch and rebased onto main (only the generated sourcemap conflicted; regenerated). 1,061 passing.

Fix 1 stays an open question — if `ERR_PNPM_MISSING_TARBALL_INTEGRITY` still reproduces for you on 1.36.0, please open an issue with the lockfile entry and the install target. Thanks for turning this around so fast, and for pushing back on the part I got wrong.

PR #428, opened 2026-08-30
fix: restore previous build after failed update

Jstn-1g:
## Summary

- rematerialize the exact previous npm version or captured GitHub commit after an update command hard-fails or times out
- preserve the user's original dependency spelling and verify the restored source identity and loadable entry
- force restored-manifest installs when package bytes changed without a manifest diff, while leaving busy and user-cancelled runs alone

## Why

`pnpm add` can replace both `package.json` and the installed package directory before returning a non-zero result. The existing failure path restored only the manifest, so a rejected or partially written build could remain on disk and run after restart.

The regression reproduced that state on current `main`: the response reported a failed update and the manifest returned to `^1.0.0`, but the installed package bytes remained at `1.2.0`.

## Verification

- focused failure/recovery regressions: 6/6
- complete flow suite: 134/134
- full suite on Node 22.22.2 and Node 24.19.0: 1,113 passed, 1 skipped on each
- typecheck, build, restart smoke, preflight, registry validation, and diff check passed
- independent patch review found no remaining blocking or material issues

### DIFF
--- a/src/routes.ts
+++ b/src/routes.ts
@@ -469,9 +469,13 @@ export function mountMarketRoutes(
    * next start still fails. Re-run pnpm install against the restored
    * manifest to rematerialize the previous build's files.
    */
-  async function rollbackUpdateBuild(name: string, manifestBefore: ProfileManifestSnapshot): Promise<{ ok: boolean; detail: string | null }> {
+  async function rollbackUpdateBuild(
+    name: string,
+    manifestBefore: ProfileManifestSnapshot,
+    rematerializeWhenManifestUnchanged = false,
+  ): Promise<{ ok: boolean; detail: string | null }> {
     const rolledBack = restoreProfileManifest(config.profile, manifestBefore, activeProfileDir)
-    if (rolledBack.length === 0) return { ok: true, detail: null }
+    if (rolledBack.length === 0 && !rematerializeWhenManifestUnchanged) return { ok: true, detail: null }
     // CI=true (the market always runs pnpm that way) turns frozen-lockfile
     // on, and the restored manifest pin now disagrees with the lockfile the
     // bad add just wrote — without the flag this restore run fails with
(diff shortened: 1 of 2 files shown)


fkysly (approved):
核过了,是真 bug、真回归测试。收。

**我在 main 上验了那个失败态。** 把你的测试留下、`src/routes.ts` 换回 main 的版本:

```
× restores the previous build when an update fails after pnpm wrote new files
  AssertionError: expected '1.2.0' to be '1.0.0'
```

manifest 已经回到 `^1.0.0`,而 `node_modules` 里躺着的还是 `1.2.0` —— 也就是接口报「更新失败」,重启之后跑的却是那份被拒绝的构建。换回你的代码就绿。

**最关键的是这段判断,而且它容易被想漏:**

> Reinstalling a restored range is not sufficient: the failed add may have advanced the lockfile to another version that still satisfies that range, so `pnpm install` can keep the rejected bytes while reporting success.

这是对的,而且是那种「看起来已经修好了」的陷阱 —— 恢复 range 再 install,pnpm 会**成功**,只是成功地保留了坏字节。所以必须钉到确切版本/commit,并在事后**回读**确认(`readInstalledVersion` / `readLockCommits` 比对,再加 `hasLoadableEntry`)。恢复完还要验证恢复本身,这层我认。

**三条不启动恢复的边界也都对:**

- `result.busy !== true` —— 我确认了 `busy` 是真字段(`src/dsh-cli.ts:964`,匹配「another desktop pnpm operation is already running」)。宿主说忙的时候再插一条 pnpm 命令只会让两边都更糟。
- 用户主动取消 —— 他们可能正想看那个中间状态,替他们改掉是越权。
- `restore` 路径本身不递归恢复。

**`finally` 里那次 `restoreProfileManifest`** 也是必要的:精确恢复用的是钉死的 target,不还原就会把用户原来的 range/tag 写法(以及浮动的 `github:` 目标和 `#path:` 选择器)悄悄换掉。这条注释写清楚了。

还有一点值得说:恢复失败时你**没有**谎报成功,而是带着 detail 往上报(`reports a failed byte rollback without claiming the previous build was restored`)。自动恢复最怕的就是失败了还说自己成功了。

55 files / 1114 passed,`npm run check` 通过,六项 CI 全绿。

PR #454, opened 2026-08-31
fix: a plugin toggle no longer erases notes, channel and region

### LINKED ISSUE #435: 插件备注重启就消失了
### 升级复测 / Retested on the latest version

- [x] 我已升级到最新版并重启，问题仍然存在 / I upgraded to the latest version, restarted, and it still happens

### dshmarket 版本 / version

1.38.0

### 运行环境 / Where are you running it

普通 dsh web（浏览器打开）/ plain dsh web in a browser

### 桌面端版本与来源 / Desktop build and version

_No response_

### 发生了什么 / What happened

我输入了插件备注，刷新还在，重启后消失

### 复现步骤 / Steps to reproduce

1、打开插件市场
2、添加备注
3、重启DSH

### 导出日志 / Exported log

[dsh-market-log.txt](https://github.com/user-attachments/files/31612237/dsh-market-log.txt)

### 操作系统 / OS

WIN11 26200.9168

fkysly:
Closes #435. Reproduced, and it turned out to be **two** defects sharing one cause: `writeMarketState` could not tell *"the caller has nothing to say about this field"* from *"the caller wants it empty"*.

## The reported one: notes vanish after a restart

@KJinxxx described it precisely — the note survives a page refresh and is gone after restarting DSH. That shape is the whole clue: the note **did** reach disk.

The note route writes through a fresh read. But `routes.ts` also holds a long-lived `marketState` read once at boot, and `refreshMarketState()` synced only `disabled` / `groups` / `groupOrder` onto it — so `marketState.notes` stayed `{}` from boot forever.

The preserve rule was:

```ts
const notes = state.notes ?? readMarketState(profileDir).notes ?? {}
```

`??` does not fire on `{}` — it is not nullish. So the next write from that stale object put the empty object back.

## The one found next to it: **any plugin toggle erased the channel and region**

Five call sites in `routes.ts` write `{ disabled, groups, groupOrder }` — everything a toggle knows about. `channel` and `region` had **no fallback at all**:

```ts
...(state.channel === undefined ? {} : { channel: state.channel }),
```

So enabling or disabling any plugin silently discarded the user's update channel and download region — both deliberate choices made in the settings card. Measured before the fix:

```
notes  : {"p":"备注"}  ✅ 保住了     ← the reported bug needs the stale object
channel: undefined     ❌ 丢了
region : undefined     ❌ 丢了
```

## Fixes, each where the confusion actually is

- **`writeMarketState`** now reads the current document and falls back to it for `channel` / `region` / `regionAuto`. Neither has an "unchoose" path — once picked they only move to another value — so an absent one always means the caller had nothing to say. An unchosen region still writes nothing, because its absence is exactly what makes the boot probe run.
- **`notes` keeps the explicit-object-wins rule**, because deleting the last note has to stay expressible. The stale copy is fixed at its source instead: `refreshMarketState` now also syncs `notes`, `channel` and `region`, so there is one story instead of two.

I deliberately did *not* make emptiness mean "preserve" — that would have fixed #435 by making notes undeletable.

## Verification

- 3 regressions in `tests/state.spec.ts`: partial write preserves everything; the last note can still be deleted; an unchosen channel/region is still written as absent rather than invented.
- **Confirmed they fail without the fix** — reverted the fallbacks and got `AssertionError: expected undefined to be 'beta'`.
- `npm test` 1153 passed; `npm run check` passed.

### DIFF
--- a/src/hot.ts
+++ b/src/hot.ts
@@ -284,24 +284,55 @@ export function readMarketState(profileDir: string): MarketState {
   }
 }
 
-/** Persist the whole market state; `disabled` is the single written key. */
+/**
+ * Persist the whole market state.
+ *
+ * Every field a caller does not carry forward is taken from disk rather than
+ * dropped. Several callers legitimately know about only one part of the
+ * state — `writeMarketState(dir, { disabled, groups, groupOrder })` appears
+ * at five call sites in routes.ts — and before #435 that shape silently
+ * erased whatever else the user had chosen:
+ *
+ * - `channel` and `region` had no fallback at all, so toggling any plugin
+ *   threw away the user's update channel and download region. Both are
+ *   deliberate choices made through the settings card, and neither has a
+ *   "clear it" path: once picked they only ever move to another value. So
+ *   an absent one always means "the caller has nothing to say", never
+ *   "the user unchose it".
+ * `notes` keeps its original rule — an explicit object wins, including an
+ * empty one, because deleting the last note has to be expressible. What made
+ * #435 lose notes was not this function but a caller: the note route wrote
+ * through a fresh read while the long-lived `marketState` in routes.ts still
+ * carried `notes: {}` from boot, and the next write from that object put the
+ * empty one back. The fix for that belongs at the call site, where the two
+ * copies are, not here — see the note route.
+ *
+ * Reading before writing costs one small JSON parse on an operation that is
+ * already doing filesystem work, and it is what makes "this function writes
+ * the whole document" safe for callers that only hold part of it.
+ */
 export function writeMarketState(profileDir: string, state: MarketState): void {
   mkdirSync(join(profileDir, HOT_DIR), { recursive: true, mode: 0o700 })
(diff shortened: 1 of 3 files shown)


Jstn-1g (changes requested):
One persisted ownership edge still needs coverage: `regionAuto` does have an explicit clear path. The manual-region route assigns `marketState.regionAuto = undefined` before calling `writeMarketState`, but `const regionAuto = state.regionAuto ?? onDisk.regionAuto` interprets that clear as omission and restores the on-disk `true`. The in-memory flag is false for the current process, but after restart the saved auto-region notice returns even though the user chose the region manually.

Please distinguish property absence from an explicitly present `undefined` for this field, for example with `Object.hasOwn(state, "regionAuto")`, and add a regression that seeds `{ regionAuto: true }`, performs the manual-selection-shaped write with `regionAuto: undefined`, and verifies the key is absent afterward. The channel/region fallback is unaffected because, as the PR notes, those fields have no clear path.

PR #457, opened 2026-08-31
fix: let manual region choice clear auto marker

Jstn-1g:
## Summary

- distinguish an omitted `regionAuto` field from an explicitly cleared one
- preserve the on-disk automatic marker for partial callers, but remove it after a manual region choice
- ignore a boot-time region probe that finishes after the user chooses a region
- prevent a disposed mount's late probe from overwriting its replacement's region or notes

## Why

This is a follow-up to #454. That merge correctly preserves fields omitted by partial state writers, but `regionAuto` has a clear path that channel and region do not: the manual-region route explicitly sets it to `undefined`. Using `state.regionAuto ?? onDisk.regionAuto` restored the previous `true`, so the automatic-region explanation returned after restart.

The same exact-head audit also held the boot probe open across a manual choice and route disposal. Without a mount-local guard, the late callback could switch a manual `global` choice back to `china`; after disposal, it could also write the old mount's captured state over the replacement mount's region and newly saved note.

Property-presence semantics preserve genuine omission while honoring the explicit clear. The probe now exits before any routing or disk mutation when a manual/configured region or disposal has made its result stale.

## Verification

- regression mutation check: the disposed-mount test fails by erasing the replacement note when the disposal guard is removed
- focused state + flow suites: 167 passed
- full suite on Node 22.22.2 and Node 24.19.0: 55 files, 1,158 passed and 1 skipped on each
- `npm run check`, preflight, registry validation, pnpm compatibility 11/11, authenticated-browser capture guard, and `git diff --check` passed
- independent exact-diff review found no blocking or material issue

### DIFF
--- a/src/hot.ts
+++ b/src/hot.ts
@@ -319,7 +319,12 @@ export function writeMarketState(profileDir: string, state: MarketState): void {
   const notes = state.notes ?? onDisk.notes ?? {}
   const channel = state.channel ?? onDisk.channel
   const region = state.region ?? onDisk.region
-  const regionAuto = state.regionAuto ?? onDisk.regionAuto
+  // Unlike channel and region, regionAuto has an explicit clear path: a
+  // manual region choice owns this field and sets it to undefined. Preserve
+  // the disk value only when the caller omitted the property entirely.
+  const regionAuto = Object.prototype.hasOwnProperty.call(state, 'regionAuto')
+    ? state.regionAuto
+    : onDisk.regionAuto
   writeFileSync(stateFile(profileDir), JSON.stringify({
     disabled: [...state.disabled],
     groups: state.groups,
--- a/src/routes.ts
+++ b/src/routes.ts
@@ -204,6 +204,7 @@ export function mountMarketRoutes(
   commandRuntime?: PluginCommandRuntime,
   agentsLookup?: AgentsLookup,
 ): () => void {
+  let disposed = false
   // An ordinary profile must resolve under DSH_HOME by the same rules as the
   // DSH CLI. A host-authoritative explicit directory (DSH Desktop) does not
   // derive a path from this display/profile name.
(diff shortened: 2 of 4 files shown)


fkysly (approved):
**这个回归是我 #454 引入的**,你诊断得完全正确。

我在那个 PR 里把 `notes` / `channel` / `region` / `regionAuto` 四个字段一视同仁地做了「缺失就从盘上回退」,理由写的是「选过之后只会换成另一个值,所以缺失永远意味着调用方没话说」。这句话对 `channel` 和 `region` 成立,**对 `regionAuto` 不成立** —— 它恰恰有一条清除路径:手动选择区域时,那条路由拥有这个字段并显式把它设成 `undefined`。于是我的 `??` 把 `true` 又捞了回来,「已为你自动选择」的说明重启后阴魂不散。

我验了你的修法,两个方向都对:

```
手动选择后 regionAuto = undefined  ✅ 清掉了
插件开关后 regionAuto = true       ✅ 保住了
           region     = china      ✅
```

`hasOwnProperty` 才是这里该用的原语 —— 区分「属性不存在」和「属性存在但值是 undefined」。我当时用 `??` 是把两者混成了一件事,而这正是 #454 本身要解决的那类混淆,我在同一个函数里又犯了一次。注释也把这个区别写清楚了,下一个人不会再踩。

## 另外两个是你自己审出来的,而且更难发现

**启动探测跨越手动选择**:用户在探测返回前手动选了 `global`,晚到的回调把它改回 `china`。

**已弃置的挂载写穿继任者**:这个尤其要紧,因为它和 #454 是同一类 —— 陈旧对象写回。旧挂载捕获的状态会盖掉新挂载的区域**和刚存下的备注**。

我按你说的把 `if (disposed || config.region !== undefined) return` 去掉复验了一次:

```
× does not let a disposed mount apply its late automatic probe
  AssertionError: expected {} to deeply equal { Object (dsh-loop) }
```

备注确实被抹掉了 —— 变异测试是真的会红,不是摆设。

55 files / 1159 passed,七项 CI 全绿。合。

PR #489, opened 2026-09-02
feat(search): split mixed-script token boundaries

duhu2000:
## Summary

- normalize adjacent Han and Latin/number text as separate search tokens
- apply the same generic normalization to queries and every indexed field
- preserve the existing same-field matching requirement and relevance weights
- add unit and real Web E2E coverage for compact vs. spaced queries

Examples:

- `MCP管理` behaves like `MCP 管理`
- `管理MCP` behaves like `管理 MCP`
- `OAuth2授权` behaves like `OAuth2 授权`

This is intentionally market-wide. It contains no plugin-specific alias, identity, score, or default-ranking override.

## Motivation

The current normalizer collapses punctuation but does not recognize script changes as token boundaries. As a result, natural mixed Chinese/Latin queries such as `MCP管理` return no results even when the same query with a space returns relevant plugins.

On a live 2,937-entry registry snapshot:

| Query | Before | After |
|---|---:|---:|
| `MCP管理` | 0 | 28 |
| `MCP面板` | 0 | 15 |
| `MCP控制台` | 0 | 2 |
| `MCP市场` | 0 | 7 |

Each optimized result set and order matched its explicitly spaced control query.

## Performance

50 warm samples over a batch of 10 queries on the same 2,937-entry snapshot:

- baseline median: 27.949 ms/batch, 2.795 ms/query
- this branch median: 29.691 ms/batch, 2.969 ms/query
- measured overhead: about 6.2%

## Validation

- `npx vitest run tests/market-data.spec.ts` — 46/46 passed
- `npm run check` — passed
- `npm run prepack` — passed
- `npm run validate:registry` — passed (2,452 bundled entries)
- `npm run test:compat` — 11/11 passed
- `npm run test:web` — 6 files, 30/30 passed
- full remaining Vitest coverage passed after isolating one pre-existing release-channel test failure

### Existing baseline failure

`tests/flows.spec.ts > release channel > re-checks immediately when the channel changes` fails because the beta dist-tag is never queried. The same failure reproduces unchanged on the separate, unmodified baseline worktree, so this PR does not mix an unrelated release-channel fix into the search change.


fkysly (approved):
收。这是 #470 那个问题的正确解法——市场级、对称作用于查询和所有索引字段、没有任何插件专属逻辑，PR 描述里也把这条写明了。

真目录（2936 条）实测：`MCP管理` 28 命中 = `MCP 管理` 28 命中，`管理MCP` 同样 28；稳定态每次按键 1.72 ms（#409 的 WeakMap 缓存吸收了字段侧的重复规范化，新增的只是查询侧一趟正则）。中英混排是这个生态的常态查询方式，这个改动值得有。1180 passed。

PR #541, opened 2026-09-08
feat: add clear controls to plugin search fields

bulingbuling688:
Plugin searches currently have no clear control. Add a `×` button to the Discover, Favorites, Themes, and Installed search fields whenever the query is nonempty, including whitespace. Clearing restores the native input's focus and removes only that tab's query, preserving the other filters and tab queries while using the existing pagination and scroll reset behavior. Disable spellcheck on all four fields.

The shared component uses a local container to focus the host Input's native element because the primitive does not forward refs. The button supports native mouse, Enter, and Space activation, has a localized accessible name and visible keyboard focus, and reserves space beside long text. Escape remains unchanged.

Closes #524.

Validation:

- Regression test observed failing on the original implementation (missing `Clear search` button), then passing after the change.
- Component suite: 156 passed, including 8 new cases for the four tabs, whitespace, focus, language changes, independent queries, and filter/page-size preservation with pagination equivalent to manually erasing the query.
- Full suite: 59 files / 1329 tests passed. `npm run check`, final `npm run typecheck`, `scripts/preflight.mjs`, `scripts/validate-registry.mjs`, `scripts/smoke-spawn.mjs`, and `git diff --check` passed. A fresh client rebuild after commit also passed `git diff --exit-code client/`.
- Real DSH `0.1.0-rc.8` and `0.1.2-alpha.2` + Chromium on Linux/WSL: 7 search/sticky E2E tests passed on each host with `DSHM_E2E_REQUIRED=1`, no skips. Covers mouse and native Enter/Space activation, focus-visible, continued typing, sticky hit targets, and all four search layouts with long text at 1200px and 640px. Measurements wait for pointer actionability after viewport resizing. Used isolated temporary DSH profiles; no trace/HAR capture.
- The final-commit [fork CI run](https://github.com/bulingbuling688/dsh-market/actions/runs/34177355760) and [upstream CI run](https://github.com/dsh-market/dsh-market/actions/runs/34177344917) passed: Ubuntu/Windows checks, pnpm compatibility, Ubuntu Web E2E on rc.8/alpha.2, and Windows Web E2E on rc.8. Each fork Web lane passed all 46 tests, including the 7 search/sticky tests. The fork's PR-only catalog guard was skipped as expected for workflow_dispatch; the upstream guard passed.
[...]

PR #582, opened 2026-09-12
fix: preserve profile and theme state when plugin activation fails

bulingbuling688:
A failed plugin import/apply could still remove disable rows and persist an enabled choice, so the next boot loaded the plugin that the toggle had rejected. Theme replacement could leave the previous theme stopped, and group toggles did not persist the same patch/carrier choices as individual toggles.

### Changes

- Classify hot-mount results as live, deferred, or failed. Commit installed-plugin enable choices only after live or explicitly deferred activation, preserving the existing non-live HTTP/ok contract.
- Share validation and persistence across individual toggles, groups, and theme selection. Stage patch/carrier changes before activation, publish state last, and retain per-member partial success for groups under the mutation lock.
- Restore touched loader entries and previous themes on failure, including owned bundle children whose loader names differ from the package name. Report restoration failures instead of claiming success.
- Compensate late file-write failures only while files still contain this operation's output. Detect intervening edits, preserve file symlinks and permission bits, and respect read-only files.
- Keep failed hot-disposal handles available for retry. Bound asynchronous disposal, entry updates, and recovery waits to 10 seconds per operation; block overlapping work on still-pending objects and handle late rejections. A timeout releases the request lock but does not cancel the underlying plugin operation.

### Compatibility and scope

Normal deferred activation remains available for unsupported hot loading, complex patches, and activation timeouts. Ordinary completed disables that leave a fiber up retain the existing restart fallback; theme replacement requires the old theme to stop.

This does not duplicate #550's module resolution or failed-load handle/temp-file cleanup, or depend on #551's desktop acknowledgement bridge. Failed loads can still leave partial runtime resources. File compensation is not crash-atomic across files and cannot undo arbitrary plugin side effects. Whole install transactions, uninstall orphan rows, and historical damaged-profile repair remain outside scope.

### Validation

- Full local unit/component suite: **1372 passed**, none skipped.
- `npm run check`, final typecheck, preflight, smoke-spawn, and diff checks passed; no dependency, version, or generated client changes.
- Real pnpm compatibility matrix: **14 passed**.
- Complete web suites on Linux/WSL, Node 24.20.0, with DSH **0.1.0-rc.8** and **0.1.5-rc.2**: **45 passed each**, none skipped (`DSHM_E2E_REQUIRED=1`).
- Five real isolated-host scenarios verify missing-export rejection with unchanged files and a serviceable restart, normal deferred activation, durable group carrier toggles, failed theme replacement recovery, and successful theme replacement across restart.
- Regressions failed on the original baseline and before the deeper fixes. Independent local mutations removing rejection, file compensation, theme restoration, strict disposal, or the lifecycle wait bound make the corresponding regressions fail again.
- Pending/late-rejection cases use the production routes and real temporary files with controlled loader promises and virtual clocks. Windows, alpha, and Desktop acknowledgement environments were not tested locally. POSIX read-only permission cases explicitly skip Windows/root.

### Code-change checklist

- [x] `npm test` and `npm run check` pass locally.
- [x] Regression tests fail without the relevant fix.
- [x] No version bump in `package.json`.
- [x] Client source unchanged; rebuild produces no `client/client.js` diff.

Refs #575

### DIFF
--- a/src/atomic-write.ts
+++ b/src/atomic-write.ts
@@ -1,0 +1,35 @@
+/** Small profile files must not be truncated by a rejected write. */
+import { accessSync, constants, lstatSync, realpathSync, renameSync, rmSync, statSync, writeFileSync } from 'node:fs'
+import { randomUUID } from 'node:crypto'
+import { dirname, join } from 'node:path'
+
+/** Follow an existing file link without replacing the user's link itself. */
+export function writeDestination(path: string): string {
+  let link = false
+  try { link = lstatSync(path).isSymbolicLink() } catch (error) {
+    if ((error as NodeJS.ErrnoException).code !== 'ENOENT') throw error
+  }
+  // A dangling link is an error, not permission to replace it with a file.
+  return link ? realpathSync(path) : path
+}
+
+/** Atomic replacement must respect the existing file's write permissions. */
+export function checkWritableFile(path: string): void {
+  try { accessSync(path, constants.W_OK) } catch (error) {
+    if ((error as NodeJS.ErrnoException).code !== 'ENOENT') throw error
+  }
+}
+
+export function writeFileAtomic(path: string, text: string): void {
+  const target = writeDestination(path)
+  checkWritableFile(target)
+  let mode = 0o600
+  try { mode = statSync(target).mode & 0o777 } catch (error) {
+    if ((error as NodeJS.ErrnoException).code !== 'ENOENT') throw error
+  }
+  const temporary = join(dirname(target), `.dshm-write-${randomUUID()}`)
+  try {
+    writeFileSync(temporary, text, { flag: 'wx', mode })
+    renameSync(temporary, target)
+  } finally { rmSync(temporary, { force: true }) }
+}
--- a/src/hot.ts
+++ b/src/hot.ts
@@ -18,6 +18,8 @@
 
 import { mkdirSync, readdirSync, readFileSync, rmSync, writeFileSync } from 'node:fs'
 import { join } from 'node:path'
+import { writeFileAtomic } from './atomic-write.ts'
+import { waitForLifecycle } from './lifecycle.ts'
 import { pathToFileURL } from 'node:url'
 import { asChannel, type Channel } from './channels.ts'
 import { asRegion, normalizeGithubProxy, type Region } from './regions.ts'
@@ -352,7 +354,7 @@ export function writeMarketState(profileDir: string, state: MarketState): void {
   const githubProxy = Object.prototype.hasOwnProperty.call(state, 'githubProxy')
     ? state.githubProxy
     : onDisk.githubProxy
-  writeFileSync(stateFile(profileDir), JSON.stringify({
+  const text = JSON.stringify({
     disabled: [...state.disabled],
     groups: state.groups,
     groupOrder: state.groupOrder,
@@ -367,7 +369,8 @@ export function writeMarketState(profileDir: string, state: MarketState): void {
     ...(region === undefined ? {} : { region }),
     ...(regionAuto === true ? { regionAuto: true } : {}),
     ...(githubProxy === undefined ? {} : { githubProxy }),
-  }))
+  })
+  writeFileAtomic(stateFile(profileDir), text)
 }
 
 /** Plugins the user switched off; skipped by the boot re-mount. */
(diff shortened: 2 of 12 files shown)


liuwenji007:
Thanks for pushing the deeper consistency work (stage-then-commit, shared path for toggles/groups/themes, failure restore, atomic writes). That direction matches the #575 class of bugs, and the test story looks serious.

Please sync with what’s already on main first: the core #575 hole is already fixed by #585 (in 1.46.1) — a failed enable no longer flips the durable patch layer, and state.json / in-memory views are restored too (#584 was closed unmerged in favor of that landing).

So this PR should not be treated as the P0 crash-loop fix anymore; it’s hardening on top of #585. It’s also not mergeable right now (dirty vs main). Please:

Rebase onto current main;
Spell out which gaps remain after #585 (e.g. theme swap failure, group persistence, write compensation);
Confirm the #575 regressions still fail without the remaining fixes and pass after rebase.
Happy to re-review the implementation once conflicts are gone and CI is green. Until then, no LGTM. If the PR description still frames this as fixing the main #575 path, please reword it as follow-up hardening beyond #585.

fkysly (changes requested):
方向我认——「失败的启用不能留下一个看起来成功的状态」是 #575 那条保证的延续，而它至今只在单插件那条路上完整成立。但这条**必须 rebase**（当前 main 上 4 处冲突），而且 rebase 之后有一半可能已经不需要了，先说清楚免得你白改。

## 我核过的一处：分组那条路已经在保护里了

```ts
// src/routes.ts:2597 —— 分组切换对每个成员走的就是同一个函数
: await setPluginEnabled(member, enabled)
```

`setPluginEnabled`（525 行）就是 #575 加保护的那个——启用失败时把内存里的 disabled 集合恢复回去、不写 state。所以「分组切换会把成员停用状态写坏」这半条，**在当前 main 上不成立**（除非你观察到的是 patch 行那条路，见下）。

请 rebase 之后重跑一遍你的复现，把**仍然复现**的那几条列出来。你这 1378 行里如果有一半是在修已经修好的东西，合进去的代价是真实的：多出来的分支没人会再读一遍。

## 我关心的三条

按重要性排：

1. **patch / carrier 行的时序**：如果分组切换先写了 patch 行、激活随后失败，那些行会不会留下？这才是你说的「stage before activation, publish state last」要解决的。**请给出一条能复现的序列**（哪几步、哪一步失败、盘上留下了什么）。
2. **主题替换把上一个主题停掉**：这个听起来最实在——失败后用户的主题处于「旧的停了、新的没起来」。请给一条「替换失败 → 界面/配置处于什么状态」的断言，而不只是「恢复成功」的断言。
3. **分组的部分成功**：成员逐个启用，中途一个失败——剩下的成员是继续还是中止？你说的「retain per-member partial success」我倾向继续 + 如实报告，但请确认这就是你实现的语义，因为两种都能自圆其说。

## 一点要求

你写了「preserving the existing non-live HTTP/ok contract」——这句很重要，请**用测试钉住它**：非 live（deferred）仍然返回 `ok: true`（那是既有的契约，客户端据此判断"需要重启"而不是"失败"），只有真正的失败才 `ok: false`。这个契约有三处在读（单插件、分组、客户端），改坏一处就是「插件其实装上了但界面说失败」——这一类我们刚在别处修过。

rebase 完叫我。

fkysly:
再催一次，只差 rebase：main 上现在 4 处冲突，而且 rebase 之后有一半可能不需要了——先说清楚免得你白改。

**已经在保护里的**：分组切换对每个成员走的就是 `setPluginEnabled(member, enabled)`，而它正是 #575 加保护的那个函数（启用失败时恢复内存里的 disabled 集合、不写 state）。所以「分组切换会把成员停用状态写坏」这半条在当前 main 上不成立。

rebase 之后请把**仍然复现**的那几条列出来，我要的是这三样（review 里写过，这里再收一下）：

1. **patch / carrier 行的时序**——分组切换先写了 patch 行、激活随后失败时，那些行会不会留下？给一条能复现的序列：哪几步、哪一步失败、盘上留下了什么。
2. **主题替换失败**——断言「替换失败后，界面和配置处于什么状态」，而不只是「恢复成功」。
3. **分组部分成功**——成员逐个启用、中途一个失败，剩下的继续还是中止？确认你实现的语义是哪种（我倾向继续 + 如实报告）。

外加一条**用测试钉住**的：非 live（deferred）仍然必须 `ok: true`（既有契约，客户端据此判断「需要重启」而不是「失败」）。这个契约有三处在读。

要是你这两周没空，回一句也行——我可以按你的设计自己落到 main 上并署你的名（#635、#551、#541、#523 已经这么处理了），你就不用管 rebase。

Issue #262, opened 2026-08-21
插件页面非常卡


ZeroOrigin64:
### 升级复测 / Retested on the latest version

- [x] 我已升级到最新版并重启，问题仍然存在 / I upgraded to the latest version, restarted, and it still happens

### dshmarket 版本 / version

1.17.1

### 运行环境 / Where are you running it

普通 dsh web（浏览器打开）/ plain dsh web in a browser

### 发生了什么 / What happened

插件市场页面滚动特别卡

<img width="1231" height="768" alt="Image" src="https://github.com/user-attachments/assets/084c9809-b36a-46e3-988d-f81a9f51288e" />

Chrome Profile 文件如下,  F12打开控制台,  切换到 Performance,  点击 Load Trace 按钮加载

<img width="583" height="70" alt="Image" src="https://github.com/user-attachments/assets/2a451324-0a01-4c55-a165-9a458f9a7301" />

[Trace-20260822T011719.json.gz](https://github.com/user-attachments/files/31314864/Trace-20260822T011719.json.gz)

### 复现步骤 / Steps to reproduce

打开设置 - 插件市场, 然后上下滚动即可复现

### 导出日志 / Exported log

[dsh-market-log.txt](https://github.com/user-attachments/files/31314840/dsh-market-log.txt)

### 操作系统 / OS

Windows 10

PR #579, opened 2026-09-11
fix(discovery): read dsh.engines.dsh as well as the top-level engines.dsh

JINITAIMEI121:
## Problem

Fixes #577: `manifestFacts()` read the host requirement only from the top-level `engines.dsh`, while a large part of the ecosystem (including `@linxin666/dsh-web-all` and several other packages) declares it as `dsh.engines.dsh` under the manifest's own `dsh` field. For those packages `enginesDsh` was always `null`, so the #404/#522 install-time preflight, the #473 filter, and the #520 catalog split all saw "no declaration".

Both shapes are in active use — this is not one side writing it wrong, it is the field position never having a single convention, and the market only reading one half.

## What changed

`src/discovery-compatibility.ts` — `manifestFacts` now reads both positions: top-level `engines.dsh` remains authoritative, and `dsh.engines.dsh` (under the manifest's own `dsh` field) is the fallback when the top-level declaration is absent. The output shape is unchanged (`NpmManifestFacts`), so the discovery cache (`discovery-compatibility-v1.json`) and every downstream consumer keep working as-is.

## Testing

`tests/discovery-compatibility.spec.ts` — three new cases in a dedicated describe block:

- nested `dsh.engines.dsh` alone → `enginesDsh` picked up (the #577 shape, with the reporter's actual range `>=0.1.5-rc.1` and whitespace normalization);
- both shapes present → the top-level declaration wins;
- neither shape → still `null`.

Full suite: 1334/1334 pass (60 files). `tsc -p tsconfig.json --noEmit` clean.

PR #578, opened 2026-09-11
fix(pnpm-compat): give ERR_PNPM_NO_MATCHING_VERSION its own code and wording

JINITAIMEI121:
## Problem

Fixes #569: a host peer that only ever published pre-releases (the issue's example: `@deepseek-ai/dsh-tools`, where every published version is `-rc.*`/`-alpha.*`) makes pnpm fail with **`ERR_PNPM_NO_MATCHING_VERSION`** — but `classifyPnpmFailure` only knows `ERR_PNPM_FETCH_404`, so the #289 one-shot `AUTO_INSTALL_PEERS_OFF` retry never fired: no retry, and the raw pnpm wall of text (whose "The latest release of … is 0.0.1-rc.1" line reads like "you picked the wrong version") was shown instead.

This also addresses the two review points fkysly left on #528 (closed with unmerged commits):

1. **Give the shape its own code and its own sentence, don't reuse fetch-404's.** The fetch-404 copy ("a dependency does not exist on the registry … it may be a ghost entry left in the profile's package.json") is wrong for this shape — the package exists; the *requested range* matches nothing.
2. **No version bump in the PR.** This PR does not touch `package.json`.

## What changed

- `src/pnpm-compat.ts`: new `no-matching-version` failure code with its own bilingual message. For `@deepseek-ai/dsh-*` peers it names the host-provided provision ("the runtime provides it, so npm carries no such version"); for other packages it says the declared range resolves to nothing. `pkg` extraction keeps #528's fallback regex (the `No matching version found for <pkg>@<range>` shape) and adds the existing `GET …/:` shape as a fallback.
- `src/install.ts`: the #289 retry gate now also accepts `no-matching-version` — `isUnpublishedHostPeer` (host-namespace + absent from the profile manifest) still decides, exactly as the review asked ("isUnpublishedHostPeer 那条重试逻辑照旧共用就行，只要把新 code 一起加进判断里").

## Testing

`tests/pnpm-compat.spec.ts` gains five cases whose fixture is **verbatim pnpm 12.4.1 output** captured from `pnpm add @deepseek-ai/dsh-tools@'>=0.1.0'` against the live registry — the exact prerelease-only-host-peer shape of #569, word-for-word including the wrapped URL, per the house rule of pinning real wording in classifier fixtures:

- own code `no-matching-version` (not `fetch-404`);
- scoped host peer extracted through the wrapped output (`@deepseek-ai/dsh-tools`);
- message names the host-provided provision and the automatic retry;
[...]

PR #607, opened 2026-09-14
perf(client): memoize local catalog matching to stop per-keystroke fu…

JINITAIMEI121:
## Problem

Every rendered card calls `matchInstalledName()`, which for **link:/file:** dependencies runs `findCatalogEntryForLocal()` **uncached** — a full catalog filter, a full URL walk, extra scans on the identity branch. One local dependency in the profile made every render (every search keystroke) cost hundreds of milliseconds (#589: ~300ms at 24 cards × 3,627 entries, linear in the number of local deps). The version-spec branch of the same hot path has been memoized since #262 (`looseMatchCountCache`); the local branch was missed.

## Fix

Memoize `findCatalogEntryForLocal()` itself — the function is pure, so the memo is invisible to behavior:

- outer `WeakMap` keyed by the **catalog array identity**: a refetched catalog is a new array, so invalidation happens for free and stale answers are impossible;
- inner key: name + identities + hints (the full input tuple), so different evidence for the same name never collides;
- **null results are cached** — they are the common answer for same-named forks without evidence, and they are exactly what the render loop repeats.

This fixes every caller (the render loop, discover badges, theme cards) rather than only the one call site. The host-side restore path benefits too, though it was never hot.

## Tests

Two specs in `tests/sources.spec.ts`:

- repeated calls against one catalog array keep answering correctly across rounds (memo transparency, nulls included);
- a refetched catalog array (new identity, new entry) is honored as fresh — no stale null served from the old array's cache.

All 13 existing `findCatalogEntryForLocal` behavior specs pass unchanged; sources + market-data suites green locally, client typecheck clean. The `client/client.js` artifact is rebuilt in CI against the merged tree (the `git diff --exit-code client/` gate verifies it).


fkysly:
同一个 issue（#589）有两条 PR 同时在改，@po-et 的 #592 早三天开出来，我合了那条。经过和理由写在下面，因为你这条也是对的，不该只收到一句「重复了」。

**你们的性能诊断完全一致，而且都对**：`looseMatchCount` 在 #262 拿到了 memo，但只覆盖版本号分支；`link:`/`file:` 走 `findCatalogEntryForLocal`，每次要把整个目录走两三趟，而且每渲染一张卡片跑一次。

**我按缓存键选的。** 他的内层键用 JSON 序列化数组，你的用分隔符拼接：

```ts
`${name}${identities.join('')}${hints.join('')}`
```

分隔符选得很谨慎（`` 不是空串），所以不会有一般意义上的拼接碰撞。但有一处它分不开：**`[]` 和 `['']`**——两者拼出来都是空串，而它们在 `findCatalogEntryForLocal` 里是**不同的问题**：`identitySet.size > 0` 决定进不进「有证据」分支，进了就拒绝猜测，不进则退回唯一同名匹配。

这在实践中大概率碰不到（真实 identity 不会是空串，路由也只在非空时写入）。我选另一条不是因为你这条有 bug，而是因为：**这个匹配器的全部意义就是 #485 那条教训——别让弱信号压过确定信号**；一个分不清「没有证据」和「有一条空证据」的键，会让它以缓存 bug 的形式复发。这种地方我更愿意要一个靠构造成立的不变量，而不是一个靠输入恰好不会长成那样成立的。

**你这条里有一个判断比他的好，我记下来了**：你把 memo 放在 `src/catalog-local-match.ts`，也就是客户端和宿主恢复路由**共用**的那个模块。位置上更对。我最后没按这个选，是因为热点路径确实在客户端——服务端那两个调用点是每次请求一遍，不是每渲染一张卡片一遍——所以共用带来的额外收益在这个 issue 上很小，不足以抵掉键的差别。

如果你愿意，把 memo 挪到共用模块、沿用已合入的 JSON 键，是一条干净的后续 PR；我会收。

另外顺带说一句：#578 和 #579 都是你的，质量都很高，#578 那条按 peer 与否拆分文案是我没要求而你做对的部分。这条没合不代表判断有问题——两个人撞题是队列长的结果，不是你的问题。

JINITAIMEI121:
Follow-up landed, per your note: the memo now stays in the shared `src/catalog-local-match.ts` module **and** uses the JSON-serialized key from the merged #592 (`JSON.stringify([name, identities, hints])`), so `[]` and `['']` serialize differently and the evidence-branch distinction can never be collapsed by a cache hit. Added a regression spec pinning exactly that: unique-name with no identity matches, the same call with `['']` refuses to guess (null), and the `[]` answer still comes back after both — the cache is invisible to that distinction.

Also swapped the delimiter-join comment to document *why* the key is JSON (the #485 lesson — weak signals must not override strong ones). The client bundle is rebuilt in CI against the merged tree.

Thanks for the detailed review — and for the pointer on the key; the collision argument is right and I'd rather not have shipped that.

PR #652, opened 2026-09-21
fix(pnpm-compat): keep the unexpected-store advice honest when the mismatch is Desktop staging

elkaix:
## Problem

DSH Desktop installs market plugins in a disposable staging workspace under `$DSH_HOME/profiles/.generations/staging/`. When that tree carries no `pnpm-workspace.yaml` of its own, pnpm resolves the workspace root by walking the ancestor chain — and a `pnpm-workspace.yaml` **above** the staging tree (a home folder used as a pnpm workspace is the usual one) claims the install together with its own store, whose pnpm major can differ from Desktop's bundled pnpm.

The result is `ERR_PNPM_UNEXPECTED_STORE` pointing at the staging tree, but the current advice tells the user to relink the **profile** directory:

> Relink by running `pnpm install --store-dir <the first path above>` once in the profile directory

That advice is wrong for this variant — the profile's `node_modules` is not the one that mismatched. Users who follow it churn their profile for nothing while the next Desktop install fails the same way.

Reproduction on macOS (`~/pnpm-workspace.yaml` + `~/node_modules` linked to store `v11`, Desktop bundling pnpm `10`): identical staging trees succeed in `/tmp`, fail under `$HOME` without a workspace file, and succeed with one. dshmarket then shows the profile-relink message above.

## Fix

`classifyPnpmFailure` extracts pnpm's named dependencies directory (`The dependencies at "…"`). When it sits under `.generations/staging` — Desktop's disposable staging root, not every `.generations` path — the message branches to the staging variant: it names the captured directory and both store paths, explains that this is not the profile's `node_modules`, deliberately drops the `--store-dir` suggestion, and lists the actual remediations (relink the outer workspace with its own pnpm major, or remove that ancestor `pnpm-workspace.yaml`). The profile variant keeps the current advice verbatim.

## Testing

- new specs: staging-path output → staging message (dir + both stores, outer-workspace remediations, no `--store-dir` / `allowBuilds` / Desktop-upgrade claim); `.generations/live/…` keeps profile-relink advice
- `tests/pnpm-compat.spec.ts` unexpected-store cases pass
- `npm run typecheck` clean (all three tsconfigs)


liuwenji007:
The diagnosis is right。 I agree this variant is not the profile's node_modules, and the current --store-dir / profile-relink advice is the wrong one.

Please don't merge the message as written. The PR body lists the honest remediations (relink the outer workspace with its own pnpm major, or remove that ancestor pnpm-workspace.yaml). The string in classifyPnpmFailure says two other things that look false in the same way this PR is trying to stop:

1. "Add an allowBuilds key to the profile's pnpm-workspace.yaml so staging always carries its own workspace file" — allowBuilds only approves build scripts. It does not create a workspace file under .generations/staging, and it does not stop an ancestor workspace from claiming the tree. Users who follow this will edit the profile for nothing.
2. "Update DSH Desktop (newer versions isolate staging)" — I cannot find a version this is true of. If you have one, name it; otherwise drop the claim.
3. "Do not relink as the general advice below suggests" — this branch returns, so there is no advice below.

Also tighten the path check to `.generations/staging`, not every `.generations`. A `.generations/live/…` mismatch is not the staging workspace, and this copy would mis-describe it.

Keep the branch; make the copy match the PR body (name the staging dir + the two stores; say it is not the profile; give the two remediations above). I'll look again after that.

liuwenji007:
The four points from the last review are all addressed. Thanks for the follow-up commit.

The path check is `.generations/staging` only (`live/` still gets the profile-relink copy). allowBuilds and the Desktop-upgrade claim are gone. The remediations match the PR body (relink the outer workspace with its own pnpm major, or remove the ancestor pnpm-workspace.yaml), and it no longer refers to advice that is not on the page.

Fine to merge. I'll take a small follow-up for the Chinese half of this string, the profile variant in the same function is bilingual, and this branch is English-only. No need to spin another commit on your side.

